PR #9, opened 2018-12-10
minor edits in cicle yml

### LINKED ISSUE #3: Create a readthedocs page
hey @chrisfilo can you create a readthedocs page for this repository? I don't think I have permissions to do so...

sappelhoff:
closes #3


sappelhoff:
weirdly enough, the docs are built properly in #10 ... although the building process is no different from master (where it fails)

.circleci/config.yml
```diff
@@ -36,16 +34,14 @@ jobs:
...
             - cache-pip
+      - run: pip freeze
       # Install the packages needed to build our documentation
-      # This will depend on your particular package!
-      - run: pip install --user -r docs/requirements.txt
-      - run: pip install --user .
+      - run:
+          name: install required packages
+          command: |
+            pip install --user -r docs/requirements.txt
```
choldgraf: hmmm - why no more `pip install --user .`? Don't the docs need pybv to build?

choldgraf: sounds good - circle cacheing is kind of a mystery to me too :-)

sappelhoff:
@choldgraf I made some changes: The build is now using conda for the management of the python environment.

- Pro: It works and it will be easy to extend!
- Con: The python environment is not transferred from the "test" job to the "build_docs" job --> this means duplication of code in the circle `config.yml`, and some more time that it needs to run
- Alternative 1: Try to manage all of this with pipenv and this guide: https://circleci.com/docs/2.0/language-python/
- Alternative 2: Move code testing to Travis CI, and use CircleCI only for docs building

I am happy with how it is in this PR, or with `Alternative 1`, if you want to give it a shot.

We also need to discuss
1. to which documentation we should link ("stable", or "latest") ... or is there a convenient way for users to switch while they are ON the readthedocs page?
2. we need to update the installation procedure to mention that `pip install pybv` is now an option to get the "stable alpha" ;-) instead of the "dev alpha"

choldgraf:
I'm +1 on using the state of this PR and seeing how it goes from there (just FYI I generally don't review things in detail if the prefix is `[WIP]`...is it ready for review?

sappelhoff:
> I'm +1 on using the state of this PR and seeing how it goes from there (just FYI I generally don't review things in detail if the prefix is [WIP]...is it ready for review?

... you are right ... I should have changed to [MRG] :-)

feel free to merge and we see how it goes.

PR #10, opened 2018-12-10
[MRG] trigger codecov

### LINKED ISSUE #8: add coverage, coverage CI, coverage Badge
e.g., https://codecov.io/

sappelhoff:
closes #8

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,8 +21,14 @@ jobs:
             - ~/.cache/pip
       # Run tests
       - run:
-          name: Running tests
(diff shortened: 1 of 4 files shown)


choldgraf:
I assume that the `n/a` in the report above is because this is the first time we've used codecov? If so I'll merge, just wanna make sure it's not a bug :-)

sappelhoff:
> I assume that the n/a in the report above is because this is the first time we've used codecov? 

yes, I think so :-)

choldgraf:
boom 🎉

PR #13, opened 2019-03-05
Let numpy.ndarray.tofile save the data

tstenner:
The [`tofile`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.tofile.html) function saves the data directly to a file. In tests with a 100MB dataset this is about 16% faster and requires at least one less copy in RAM.


choldgraf:
nice!

this seems like a more elegant way to do it to me, @sappelhoff what do you think?

jasmainak:
+1 I've seen `np.fromfile` being used in a lot of MNE IO code.

sappelhoff:
Thanks!

PR #14, opened 2019-04-18
Add missing comment marker

tstenner:
Some readers can't handle the uncommented comment otherwise

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -155,7 +155,7 @@ def _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq, ch_names,
             print(r'DataFormat=BINARY', file=fout)
 
         if 'multiplexed' == orientation.lower():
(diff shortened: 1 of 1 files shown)


sappelhoff:
Mmh could you specify (i) which readers, and (ii) what you mean by "uncommented comment"?

according to the [BrainVision Core Data Format](https://www.brainproducts.com/files/public/products/more/BrainVisionCoreFileFormat_1-0.pdf), `DataOrientation` is a proper field, not a comment.

Although I see that we currently have `Data orientation` with a space in between ... :thinking:

tstenner:
The first line is a comment that describes what multiplexed means in general, the second line is what should be read:

```
; Data orientation: MULTIPLEXED=ch1,pt1, ch2,pt1 ...
DataOrientation=MULTIPLEXED
```

Some readers ignore the first line even if it's not a comment, but some others (e.g. the EEGLAB reader) can't handle it.

sappelhoff:
Thanks for the explanation and PR to fix it :+1:

PR #16, opened 2019-04-24
Prepare for release

sappelhoff:
After @tstenner has contributed some fixes and improvements, I think we can release a very modest `0.0.2` version of pybv.

I took this opportunity to also make some minor fixes and add a "how to release" section to our CONTRIBUTING.md

If you approve, I'll go ahead and merge this and then make the release.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,15 +45,24 @@ install the following:
 Then from the command line navigate to the `/docs` directory and call `make
 html`. The HTML pages will be in `_build/html`.
 
-##### Making a release (needs admin rights)
+##### Making a release
+(needs admin rights for GitHub and PyPi)
+
 1. go to your python environment for `pybv`
 1. run `pip install -U setuptools wheel twine`
 1. update the `__version__` variable in `__init__.py`
-1. commit the change and then git tag the commit with the version
-1. git push to master (or make a pull request)
-1. also push the git tag to upstream: `git push upstream --tags`
-1. run `python setup.py sdist bdist_wheel`
-1. run `twine upload dist/*` (needs admin rights for pypi)
-1. Make a release on GitHub
-1. append `-dev` to the `__version__` variable in `__init__.py` (updating it
-   accordingly)
+    - remove the `-dev` suffix
+    - bump up the `major.minor.patch` version according to
+      [semantic versioning](https://semver.org/)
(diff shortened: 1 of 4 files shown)


tstenner:
If it's not urgent, I'd like to add another PR for more control over the saved resolution / scale before the next release comes out.

sappelhoff:
> If it's not urgent, I'd like to add another PR for more control over the saved resolution / scale before the next release comes out.

Fine with me, what's your timeline? If it's more than two weeks we can still release now, and make another release once your pr is added in the future

tstenner:
Then it's less than two weeks :-)

CONTRIBUTING.md
```diff
@@ -44,3 +44,16 @@ install the following:
 
 Then from the command line navigate to the `/docs` directory and call `make
 html`. The HTML pages will be in `_build/html`.
+
+##### Making a release (needs admin rights)
+1. go to your python environment for `pybv`
+1. run `pip install -U setuptools wheel twine`
+1. update the `__version__` variable in `__init__.py`
```
choldgraf: do we want to adopt something like `<version>.dev0` for pybv? I've found it makes it easier to keep track of when we're in a dev version vs. when we need to bump versions etc.

CONTRIBUTING.md
```diff
@@ -44,3 +44,16 @@ install the following:
...
+1. commit the change and then git tag the commit with the version
```
choldgraf: I believe that if you create a release via the github UI, it'll also tag it, I've found that to be more intuitive for folks

CONTRIBUTING.md
sappelhoff: +1, we are currently using `-dev`, see [here](https://github.com/bids-standard/pybv/blob/f1c84c062b8f1e3596d63b331c5d205d5b9c109e/pybv/__init__.py#L10)

CONTRIBUTING.md
sappelhoff: That's a good suggestion, thanks!

tstenner:
PR is submitted.

sappelhoff:
of course I realize this one minute AFTER the release ....

#17 was actually a new feature so according to semver, we should be at `0.1.0` and not at `0.0.2`. Oups :-(

well I hope nobody gets hurt.

PR #17, opened 2019-04-27
Support channel-specific scaling factors

### LINKED ISSUE #12: Support channel-specific scaling factors
The brainvision format allows each channel to have its own scaling factor, e.g.

```
; Each entry: Ch<Channel number>=<Name>,<Reference channel name>,<Resolution in "Unit">,<Unit>, Future extensions..
Ch1=Fp1,,0.2,µV
Ch2=Fp2,,6.5,nV
```

This is really useful with int16 as datatype since it maximizes each channel's range without impairing precision too much (and it's also what BrainVision Analyzer does).

The only required (backwards compatible) change in the interface would be in the `resolution` parameter, which could either accept `float` array with `nchan` elements or a single `float` that gets promoted automatically.

tstenner:
closes #12

This changes the resolution parameter to accept either a single parameter or an 1xnchan array for channel specific resolutions, adds a `scale_data` parameter (defaults to True) indicating whether the data needs to be scaled or is already scaled correctly and improves the scaling factor precision by allowing other units (`V` and `nV` for now).

To test it, I created a sample file with `write_brainvision(np.ones((4,12), 5, ['Foo','Bar','Baz','Fob'], 'test', '/tmp', events=None, resolution=[1e-8, 1e-5, .1, 2.5e-7], True)`, read it in with mne-python and confirmed the data array was all ones again, even though the channel resolutions were different.

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -160,7 +160,7 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events):
 
 
 def _optimize_channel_unit(resolution):
-    """Calculate an optimal channel scaling factor and unit"""
+    """Calculate an optimal channel scaling factor and unit."""
     exp = np.log10(resolution)
     if exp <= -8:
         return resolution / 1e-9, 'nV'
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -16,11 +16,11 @@ from . import __version__
 
 # ascii as future formats
 supported_formats = {
-    'binary_float32' : 'IEEE_FLOAT_32',  # noqa: E203
-    'binary_int16'   : 'INT_16',  # noqa: E203
+    'binary_float32' : ('IEEE_FLOAT_32', np.float32),  # noqa: E203
+    'binary_int16'   : ('INT_16', np.int16),  # noqa: E203
 }
 
-supported_orients = set(['multiplexed'])
+supported_orients = {'multiplexed'}
 
 
 def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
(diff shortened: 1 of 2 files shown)


pybv/io.py
```diff
@@ -70,27 +77,56 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
...
+        errmsg = ('Data format {} not supported.'.format(format) +
+                  'Currently supported formats are: ' +
+                  ', '.join(supported_formats))
+        raise ValueError(errmsg)
+    return supported_formats[fmt]
+
+
+def _chk_multiplexed(orientation):
+    """Check if the passed orientation is supported and return if
+    it is multiplexed (True) or vectorized (False)"""
```
sappelhoff: Please revise docstring according to [pep257](https://www.python.org/dev/peps/pep-0257/):

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

pybv/io.py
```diff
@@ -70,27 +77,56 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
...
     _write_vmrk_file(vmrk_fname, eeg_fname, events)
     _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq,
                      ch_names, resolution=resolution)
-    _write_bveeg_file(eeg_fname, data, resolution=resolution)
+    _write_bveeg_file(eeg_fname, data, resolution=resolution,
+                      scale_data=scale_data)
+
+
+def _chk_fmt(fmt):
+    """Check that the format string is valid, return BVEF / numpy datatypes"""
```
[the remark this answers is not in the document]
tstenner: Done.

tstenner:
> Could you add this test, please?

Also done, I modified the existing test to write the 10 channels in resolutions ranging from 1V to .1nV and read it back in.

pybv/io.py
```diff
@@ -124,23 +159,25 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events):
                   .format(ii, tformat.format(i_val), i_ix), file=fout)
 
 
+def _optimize_channel_unit(resolution):
+    """Calculate an optimal channel scaling factor and unit"""
```
sappelhoff:
```suggestion
    """Calculate an optimal channel scaling factor and unit."""
```

pybv/tests/test_bv_writer.py
```diff
@@ -105,3 +106,13 @@ def test_bv_writer_oi_cycle():
     assert ch_names == raw_written.ch_names
 
     rmtree(tmpdir)
+
+
+def test_scale_data():
+    """Test that scale_data=False keeps the data untouched"""
```
sappelhoff:
```suggestion
    """Test that scale_data=False keeps the data untouched."""
```

pybv/tests/test_bv_writer.py
```diff
@@ -105,3 +106,13 @@ def test_bv_writer_oi_cycle():
...
+
```
sappelhoff:
```suggestion
```
removing superfluous empty line

sappelhoff:
Thanks @tstenner!

PR #20, opened 2019-04-29
Increase effective resolution

tstenner:
For some adventurous resolutions, the defaults are inadequate. This PR tweaks them so even resolutions such as `9e-9` are adequately written to the header file.


sappelhoff:
seems like a formatting change only?

tstenner:
Mostly, it changes the cutoff point at which the resolution switches to nanovolts and increases the number of digits for the resolution.

sappelhoff:
thanks @tstenner :-)

PR #21, opened 2019-04-29
docs: add API andChangelog, unify README, correct versions ... tests: fix coverage

### LINKED ISSUE #18: update usage example in README
... to account for new parameters that the `write` function can take

https://github.com/bids-standard/pybv/blob/master/README.md

### LINKED ISSUE #19: turn README.md to RST and input into docs index via "include"
This would make it easier for us to keep consistent data between the README and our documentation --> only one place to maintain both.

for the include directive, see an example here: https://github.com/mne-tools/mne-bids/blob/master/doc/index.rst

a nice md to rst converter: https://github.com/miyakogi/m2r

sappelhoff:
big maintenance PR :-)

closes #18 
fixes #19

- fix versioning and the guide (see https://github.com/bids-standard/pybv/pull/21#issuecomment-488644785)
- Unify README.md and index.rst --> transform into a single README.rst ... and "include" it in index.rst
- Fix the coverage reporting CI (it was broken for a long time and we did not notice)
- Use flake8 in CI to check for pep8 compliance of the code
- add a changelog and an API to the documentation
- prune the configuration file for sphinx from defaults and unnecessary options


sappelhoff:
I'll also fix our versioning scheme in here: currently we append `-dev` to the development version.

But according to [PEP440](https://www.python.org/dev/peps/pep-0440/#version-scheme), the version should be fitting `[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`

so we should append `.dev0` instead of `-dev` --> `0.2.0.dev0`

choldgraf:
This all looks like a nice improvement to me! I like the re-org.

Something that I've started to use in other projects, might be useful here, is this [ruby-based changelog generator](https://github.com/github-changelog-generator/github-changelog-generator). It's a CLI that automatically generates changelogs based on github tags and releases. Might help save some time and produces pretty nice-looking changelogs (e.g., see here https://github.com/jupyter/jupyter-book/blob/master/CHANGELOG.md)

That said, I think that could be a different PR...are you ready to merge this one?

sappelhoff:
> That said, I think that could be a different PR...are you ready to merge this one?

ready to merge - I agree: different PR would be better :-) 

but looks interesting, I'll check it out!

sappelhoff:
actually, ... it seems like we are using the same on in `bids-specification`: https://github.com/bids-standard/bids-specification/blob/ddedf0c9ae8174b04bde2e406069a59dc85c8f20/.circleci/config.yml#L33

choldgraf:
boom! done

PR #22, opened 2019-04-30
Add the format parameter to the public API

tstenner:
One functionality that's still missing from the public API is the format the data is written as (float32 or int16).

This PR adds the parameter.


tstenner:
Long story, in short we've got a lot of data in a 24bit binary format (nV resolution) we need to import in Analyzer, but most files only have an absolute voltage range of about 2.5mV so 16bit is more than enough.

sappelhoff:
Thanks, I'll add a test in my misc PR #21

PR #23, opened 2019-05-31
speedup CI build through caching

sappelhoff:
This PR was intended to speed up the CI through caching ... but I noticed that we are running into an error with our tests when using MNE 0.18.

For now I pinned MNE to 0.17, and opened an issue, that we'll have to deal with this. https://github.com/bids-standard/pybv/issues/24

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,12 +1,18 @@
 version: 2
 jobs:
-  tests:
+  run_tests:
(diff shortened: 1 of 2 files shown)


choldgraf:
This seems reasonable to me - tests are happy so I'm gonna give it a merge and we can see if the cacheing makes a noticeable improvements in subsequent PRs (and good catch on the bug)

PR #26, opened 2019-07-18
Update MNE compatibility to >0.17, fix bug with stimulus integer codes <length 3 in events

### LINKED ISSUE #24: Update to MNE 0.18 and prepare for MNE 0.19
Our tests fail with MNE 0.18:

```Shell
___________________________ test_bv_writer_oi_cycle ____________________________

    def test_bv_writer_oi_cycle():
        """Test that a write-read cycle produces identical data."""
        tmpdir = _mktmpdir()
    
        # Write, then read the data to BV format
        write_brainvision(data, sfreq, ch_names, fname, tmpdir, events=events,
                          resolution=np.power(10., -np.arange(10)))
        raw_written = mne.io.read_raw_brainvision(op.join(tmpdir, fname + '.vhdr'),
                                                  preload=True, stim_channel=False)
        # Delete the first annotation because it's just marking a new segment
        raw_written.annotations.delete(0)
        # Convert our annotations to events
        events_written, event_id = mne.events_from_annotations(raw_written)
    
        # sfreq
        assert sfreq == raw_written.info['sfreq']
    
        # Event timing should be within one index of originals
        assert_allclose(events[:, 0], events_written[:, 0], 1)
>       assert_array_equal(events[:, 1], events_written[:, 2])
E       AssertionError: 
E       Arrays are not equal
E       
E       Mismatch: 100%
E       Max absolute difference: 10000
E       Max relative difference: 0.99990001
E        x: array([1, 1, 2, 2])
E        y: array([10001, 10001, 10002, 10002])

pybv/tests/test_bv_writer.py:97: AssertionError
```

and in MNE 0.19.dev0, the stim_channel will be gone: https://github.com/mne-tools/mne-python/pull/6348/files

--> which makes this test fail as well:

```Shell
=================================== FAILURES ===================================
___________________________ test_bv_writer_oi_cycle ____________________________

    def test_bv_writer_oi_cycle():
        """Test that a write-read cycle produces identical data."""
        tmpdir = _mktmpdir()
    
        # Write, then read the data to BV format
        write_brainvision(data, sfreq, ch_names, fname, tmpdir, events=events,
                          resolution=np.power(10., -np.arange(10)))
        raw_written = mne.io.read_raw_brainvision(op.join(tmpdir, fname + '.vhdr'),
>                                                 preload=True, stim_channel=False)
E       TypeError: read_raw_brainvision() got an unexpected keyword argument 'stim_channel'

pybv/tests/test_bv_writer.py:86: TypeError

```

sappelhoff:
fixes #24

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -160,6 +160,7 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events):
         # This is a string staring with "S" and followed by an integer of
         # minimum length 3, padded with "space" if the integer is < length 3.
         # For example "S  1", "S 23", "S345"
+        # XXX: see https://github.com/bids-standard/pybv/issues/24#issuecomment-512746677  # noqa: E501
         twidth = int(np.ceil(np.log10(np.max(events[:, 1]))))
         twidth = twidth if twidth > 3 else 3
         tformat = 'S{:>' + str(twidth) + '}'
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -156,8 +156,12 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events):
         if events is None or len(events) == 0:
             return
 
-        # Handle events
+        # Handle events: We write all of them as "Stimulus" events for now.
+        # This is a string staring with "S" and followed by an integer of
+        # minimum length 3, padded with "space" if the integer is < length 3.
+        # For example "S  1", "S 23", "S345"
         twidth = int(np.ceil(np.log10(np.max(events[:, 1]))))
+        twidth = twidth if twidth > 3 else 3
         tformat = 'S{:>' + str(twidth) + '}'
 
         for ii, irow in enumerate(range(len(events)), start=2):
--- a/README.rst
+++ b/README.rst
@@ -104,8 +104,7 @@ Here is an example of the MNE code required to read in BrainVision data:
     import mne
 
     # Import the BrainVision data into an MNE Raw object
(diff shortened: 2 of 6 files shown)


pybv/io.py
```diff
@@ -156,8 +156,12 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events):
         if events is None or len(events) == 0:
             return
 
-        # Handle events
+        # Handle events: We write all of them as "Stimulus" events for now.
+        # This is a string staring with "S" and followed by an integer of
+        # minimum length 3, padded with "space" if the integer is < length 3.
+        # For example "S  1", "S 23", "S345"
         twidth = int(np.ceil(np.log10(np.max(events[:, 1]))))
+        twidth = twidth if twidth > 3 else 3
```
choldgraf: what are the reasons that it would be greater than 3?

sappelhoff: I think there are none (waiting for a reply by BrainProducts, whether this would even be supported)

but until we know for sure, I think it's better so only fix what is broken :-)

choldgraf: makes sense to me - maybe we should add in a comment saying as much? I'm always nervous about adding in these kinds of lines where it's not explicit why "3" is the right number here. Right now we know this is the case, but I imagine a future where we look at this line of code and are like "why the hell did we cap it to 3?!"

sappelhoff: sounds good :-) fixed in https://github.com/bids-standard/pybv/pull/26/commits/1baee691aa9379b550cf4fd6679471a3b5dae14b

choldgraf:
nice - LGTM then :-)

PR #29, opened 2019-07-21
FIX+ENH: Fix indexing in events and add measurement date option to writer

### LINKED ISSUE #27: potential indexing issue
Could it be that we have an indexing issue in our code?

BrainVision uses 1-based indexing for writing the events, whereas in Python and MNE-Python, we use 0-based indexing.

Now imagine a user passing `events`: https://github.com/bids-standard/pybv/blob/cf8aa9d6904149d68ad9d9500315db5e785664a2/pybv/io.py#L46-L50

These events will most likely be 0-indexed.

However, later down in the code -when we write a VMRK file, we translate these indices one to one, disregarding that BrainVision used 1-based indexing. See: https://github.com/bids-standard/pybv/blob/cf8aa9d6904149d68ad9d9500315db5e785664a2/pybv/io.py#L163-L167

Later on in the testing, we actually see this issue: https://github.com/bids-standard/pybv/blob/cf8aa9d6904149d68ad9d9500315db5e785664a2/pybv/tests/test_bv_writer.py#L95-L96

note the comment. I may be wrong ... but it seems to me we should handle the indexing better

cc @tstenner @choldgraf

sappelhoff:
This PR does two things

# 1
This PR exposes a `meas_date` argument to the writer function and a test suite. The default is `None` and maintains the current behavior.

# 2
This PR fixes #27 

I am adding `1` to each sample index, to account for BrainVision's 1-based indexing (whereas we expect an array that is 0-based indexed). Note, I changed the docstring for `events`. It now mentions that we expect a zero-based indexed array of events.

Currently the reasoning that BrainVision uses 1-based indexing is based on inspecting files, and realizing that the "New Segment" marker is always the first one, and always at sample `1` ... not at `0`

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -11,6 +11,8 @@
 import codecs
 import os
 import os.path as op
+import datetime
+
 import numpy as np
 
 from pybv import __version__
@@ -64,8 +66,9 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
     fmt : str
         Binary format the data should be written as. Valid choices are
         'binary_float32' (default) and 'binary_int16'.
-    meas_date : str | None
-        The measurement date of the data specified as a string in the format:
+    meas_date : datetime.datetime | str | None
+        The measurement date of the data specified as a datetime.datetime
+        object. Alternatively, can be a string in the format:
         "YYYYMMDDhhmmssuuuuuu". "u" stands for microseconds. If None, defaults
         to '00000000000000000000'.
     """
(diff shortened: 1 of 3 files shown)


pybv/io.py
```diff
@@ -149,9 +163,10 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events):
         print(r'[Marker Infos]', file=fout)
         print(r'; Each entry: Mk<Marker number>=<Type>,<Description>,<Position in data points>,', file=fout)  # noqa: E501
         print(r';             <Size in data points>, <Channel number (0 = marker is related to all channels)>', file=fout)  # noqa: E501
+        print(r';             <Date (YYYYMMDDhhmmssuuuuuu)>', file=fout)
         print(r'; Fields are delimited by commas, some fields might be omitted (empty).', file=fout)  # noqa: E501
         print(r'; Commas in type or description text are coded as "\1".', file=fout)  # noqa: E501
-        print(r'Mk1=New Segment,,1,1,0,0', file=fout)
+        print(r'Mk1=New Segment,,0,1,0,{}'.format(meas_date), file=fout)
```
sappelhoff: Note here: I changed the index of the "New Segment" marker from `1` to `0`, because after inspecting some VMRK files, it seems that BrainVision does zero-based-indexing after all ... There is no obvious documentation right now, so I'll have to write an email again.

pybv/io.py
```diff
@@ -64,6 +64,10 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
     fmt : str
         Binary format the data should be written as. Valid choices are
         'binary_float32' (default) and 'binary_int16'.
+    meas_date : str | None
+        The measurement date of the data specified as a string in the format:
+        "YYYYMMDDhhmmssuuuuuu". "u" stands for microseconds. If None, defaults
+        to '00000000000000000000'.
```
choldgraf: why not allow for [ISO 8601 formatting](https://en.wikipedia.org/wiki/ISO_8601) which is a bit more human-readable?

pybv/tests/test_bv_writer.py
```diff
@@ -92,8 +111,8 @@ def test_bv_writer_oi_cycle():
     # sfreq
     assert sfreq == raw_written.info['sfreq']
 
-    # Event timing should be within one index of originals
-    assert_allclose(events[:, 0], events_written[:, 0], 1)
+    # Event timing should be exactly the same
+    assert_array_equal(events[:, 0], events_written[:, 0])
```
choldgraf: 👍

pybv/io.py
sappelhoff: good point, I am now allowing `datetime.datetime` objects, I think this should cater to all needs :-)

See: https://github.com/bids-standard/pybv/pull/29/commits/d0467c2d087b77fc453de83094135c19902644ba

sappelhoff:
Thanks for the review @choldgraf All tests pass, I am merging this now and preparing a new release :-)

PR #32, opened 2019-08-21
Add option to disable writing a meas_date event

### LINKED ISSUE #31: Exported file contains additional event
When exporting a file with events, the resulting .vmrk file has one additional event with the measurement date. This is not ideal because when I read the file back in, I suddenly have one additional event as compared to the original data. Isn't there another way to store the measurement date? If not, could we add an option to disable storing the additional event (e.g. use `meas_date=None` for this purpose instead of writing a zero date, or define a new value such as `meas_date="none"`)?

cbrnr:
Fixes #31.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,7 +21,7 @@ Changelog
 
 Authors
 ~~~~~~~
-- `Clemens Brunner`_
+
 
 0.1.0
 =====
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,7 +19,6 @@ Changelog
 ~~~~~~~~~
 - Add option to disable writing a meas_date event (which is also the new default) by `Clemens Brunner`_ (`#32 <https://github.com/bids-standard/pybv/pull/32>`_)
 
-Authors
 
 
 0.1.0
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,6 +15,14 @@ Here we list a changelog of pybv.
 Current
 =======
 
+Changelog
+~~~~~~~~~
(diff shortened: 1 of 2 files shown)


cbrnr:
This is ready for merge (if you are OK with changing the meaning of `meas_date=None`).

sappelhoff:
Looks good to me, also that the meaning of "None" is changed.

Can you please document this in https://github.com/bids-standard/pybv/blob/master/docs/changelog.rst ?

After that, +1 to merge

cbrnr:
Alright, done (not sure if I got the authors list correct because I assumed you list only the authors contributing to a specific release - if all authors should be listed LMK).

sappelhoff:
We are adding the "authors" section only with the release commit :-) so you can leave that empty for now and prior to the next release, we'll update it

cbrnr:
Understood - ready!

docs/changelog.rst
```diff
@@ -15,6 +15,14 @@ Here we list a changelog of pybv.
 Current
 =======
 
+Changelog
+~~~~~~~~~
+- Add option to disable writing a meas_date event (which is also the new default) by `Clemens Brunner`_ (`#32 <https://github.com/bids-standard/pybv/pull/32>`_)
+
+Authors
+~~~~~~~
```
sappelhoff:
```suggestion

```

docs/changelog.rst
```diff
@@ -15,6 +15,14 @@ Here we list a changelog of pybv.
 Current
 =======
 
+Changelog
+~~~~~~~~~
+- Add option to disable writing a meas_date event (which is also the new default) by `Clemens Brunner`_ (`#32 <https://github.com/bids-standard/pybv/pull/32>`_)
+
+Authors
```
sappelhoff:
```suggestion

```

sappelhoff:
thanks @cbrnr !

PR #37, opened 2019-09-04
PKG: Include LICENSE in sdist

hoechenberger:


sappelhoff:
Could you explain your reasoning for this PR @hoechenberger ?

cbrnr:
See here: https://github.com/cbrnr/mnelab/pull/30

They require a license to be packaged.

hoechenberger:
Hi @sappelhoff, sorry, should have done so from the beginning. Currently, the license file is not included in the sdist, as "some" would consider good practice. I have created a `conda-forge` package of `pybv`, and the `conda-forge` folks actually _insist_ on a license file being distributed with the package, and request that one files a PR to do so upstream if that's not the case yet, like @cbrnr said. Would be nice if you could merge this PR if nothing speaks against it :)

cbrnr:
Thanks @hoechenberger!

PR #39, opened 2020-03-30
ENH: Add unit parameter to write_brainvision

### LINKED ISSUE #38: Set unit to microvolts
It seems like it is currently not possible to set the resolution and unit independently. As soon as `resolution <= -7` (which is `True` for the default), the unit is set to nV.

However, I just had a conversation with a BrainProducts developer regarding potential issues with the unit. Specifically, he said that BrainVision Analyzer currently expects the unit to be µV. If it's something else, auto-scaling doesn't work (at least the initial view).

Therefore, I would like to suggest to add an option that controls the unit. Possible values could include `'V'`, `'mV'`, `'uV'`, `'µV'`, and `'nV'`. I would set the default to `'uV'` (which is also the expected unit if the unit field in the exported file is empty).

cbrnr:
Fixes #38.

Test with the following code snippet (this could be included as a proper test):

```python
from pathlib import Path
import numpy as np
import mne
from mne.io import read_raw_brainvision
import pybv


rng = np.random.default_rng(42)
data = rng.normal(loc=0, scale=1e-5, size=(10, 10000))  # typical EEG scale
info = mne.create_info(10, 256, "eeg")
raw1 = mne.io.RawArray(data, info)

fpath = "."
fname = "export"
resolution = 1e-7
unit = "µV"  # "V", "mV", "uV", "µV", "nV", "auto"

pybv.write_brainvision(data=raw1.get_data(),  # always in V
                       sfreq=raw1.info["sfreq"],
                       ch_names=raw1.info["ch_names"],
                       fname_base=fname,
                       folder_out=fpath,
                       events=None,
                       resolution=resolution,
                       scale_data=True,
                       meas_date=None,
                       unit=unit)

raw2 = read_raw_brainvision(Path(fpath) / (fname + ".vhdr"), preload=True)

print(np.allclose(raw1.get_data(), raw2.get_data()))
print(np.max(np.abs(raw1.get_data() - raw2.get_data())))
```

This is still WIP because the following issues must be resolved:

1. Currently this does not work for `unit = 'mV'` - the checks at the end return `False` and a maximum difference of 0.05 (instead of around 1e-12 for all other cases). Any ideas?
2. The position of the `unit` parameter at the end is not ideal, if we can sacrifice backwards compatibility I'd move it directly after `resolution`.
```

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -28,15 +28,15 @@ supported_orients = {'multiplexed'}
 
 
 def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
-                      events=None, resolution=1e-7, scale_data=True,
-                      fmt='binary_float32', meas_date=None, unit='µV'):
+                      events=None, resolution=1e-7, unit='µV', scale_data=True,
+                      fmt='binary_float32', meas_date=None):
     """Write raw data to BrainVision format.
 
     Parameters
     ----------
     data : ndarray, shape (n_channels, n_times)
-        The raw data to export. Data is assumed to be in
-        **volts**. The data will be stored in **microvolts**.
+        The raw data to export. Data is assumed to be in **volts** and will be
+        stored as specified in ``unit``.
     sfreq : int | float
         The sampling frequency of the data
     ch_names : list of strings, shape (n_channels,)
@@ -54,12 +54,13 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
         third column specifies the length of each event (default 1 sample).
     resolution : float | ndarray
         The resolution **in volts** in which you'd like the data to be stored.
-        By default, this will be 1e-7, or .1 microvolts. Since data is stored
-        in microvolts, the data will be multiplied by the inverse of this
-        factor, and all decimals will be cut off after this. So, this number
-        controls the amount of round-trip resolution you want.
-        This can be either a single float for all channels or an array with
-        nchan elements.
+        By default, this will be 1e-7, or .1 microvolts. This number controls
+        the amount of round-trip resolution. This can be either a single float
+        for all channels or an array with nchan elements.
+    unit : str
+        The unit of the exported data. This can be one of 'V', 'mV', 'µV' (or
+        equivallently 'uV') , 'nV' or 'auto'. If 'auto', a suitable unit based
+        on the selected resolution is chosen automatically.
     scale_data : bool
         Boolean indicating if the data is in volts and should be scaled to
         `resolution` (True), or if the data is already in the previously
@@ -74,10 +75,6 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
         ('u' stands for microseconds). Note that setting a measurement date
         implies that one additional event is created in the .vmrk file. To
         prevent this, set this parameter to None (default).
-    unit : str
-        The unit of the exported data. This can be one of 'V', 'mV', 'uV',
-        'µV', 'nV' or 'auto'. If 'auto', a suitable unit based on the selected
-        resolution is chosen automatically.
     """
     # Create output file names/paths
     if not op.isdir(folder_out):
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -57,10 +57,10 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
         By default, this will be 1e-7, or 0.1 µV. This number controls the
         amount of round-trip resolution. This can be either a single float for
         all channels or an array with n_channels elements.
-    unit : str
+    unit : str | None
         The unit of the exported data. This can be one of 'V', 'mV', 'µV' (or
-        equivalently 'uV') , 'nV' or 'auto'. If 'auto', a suitable unit based
-        on the selected resolution is chosen automatically.
+        equivalently 'uV') , 'nV' or None. If None, a suitable unit based on
+        the selected resolution is chosen automatically.
     scale_data : bool
         Boolean indicating if the data is in volts and should be scaled to
         `resolution` (True), or if the data is already in the previously
@@ -208,7 +208,7 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date):
 def _optimize_channel_unit(resolution, unit):
     """Calculate an optimal channel scaling factor and unit."""
     exp = np.log10(resolution)
-    if unit == 'auto':
+    if unit is None:
         if exp <= -7:
             return resolution / 1e-9, 'nV'
         elif exp <= -2:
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pybv/io.py [hunk 1]
```diff
@@ -74,6 +74,10 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
         ('u' stands for microseconds). Note that setting a measurement date
         implies that one additional event is created in the .vmrk file. To
         prevent this, set this parameter to None (default).
+    unit : str
+        The unit of the exported data. This can be one of 'V', 'mV', 'uV',
+        'µV', 'nV' or 'auto'. If 'auto', a suitable unit based on the selected
+        resolution is chosen automatically.
```
sappelhoff: maybe mention that µV and uV are intended as synonyms with uV being the ASCII representation of µV

pybv/io.py
```diff
@@ -29,7 +29,7 @@
 
 def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
                       events=None, resolution=1e-7, scale_data=True,
-                      fmt='binary_float32', meas_date=None):
+                      fmt='binary_float32', meas_date=None, unit='µV'):
```
sappelhoff: look at the "parameters" section in the docstring as well:

```
    data : ndarray, shape (n_channels, n_times)
        The raw data to export. Data is assumed to be in
        **volts**. The data will be stored in **microvolts**.

```

That will need to be edited as well.

cbrnr: OK, that wasn't even true before the change, I'll update it.

pybv/io.py
cbrnr: OK!

cbrnr:
Only the "mV" issue remains. I didn't find anything in the comment you mentioned so I guess I'll have to dig a bit deeper.

cbrnr:
Found it, it's because https://github.com/mne-tools/mne-python/blob/master/mne/io/brainvision/brainvision.py#L504. I guess I'll submit a new PR there and we'll wait for it to get merged.

cbrnr:
Ready for review - all tests should pass once https://github.com/mne-tools/mne-python/issues/7537 is merged.

pybv/tests/test_bv_writer.py
```diff
@@ -137,3 +133,16 @@ def test_scale_data():
     data_written = np.fromfile(tmpdir + '/' + fname + '.eeg', dtype=np.float32)
     assert_allclose(data_written, data.T.flatten())
     rmtree(tmpdir)
+
+
+@pytest.mark.parametrize("resolution", np.logspace(-3, -9, 7))
+@pytest.mark.parametrize("unit", ["V", "mV", "uV", "µV", "nV"])
```
sappelhoff: Why `np.logspace(-3, -9, 7)`? It is of len==7, while the unit list is of len==5

pybv/tests/test_bv_writer.py
```diff
@@ -21,16 +21,16 @@
...
 from pybv.io import write_brainvision, _write_bveeg_file, _write_vhdr_file
 
-# Create data we'll use for testing
+# create testing data
 fname = 'pybv'
 np.random.seed(1337)
 n_chans = 10
-ch_names = ['ch_{}'.format(ii) for ii in range(n_chans)]
-sfreq = 1000.
+ch_names = ['ch_{}'.format(i) for i in range(n_chans)]
```
sappelhoff: I was once told that single letter variables are not so great, because they are hard to search for in code ... that's why I started to use `ii` instead of `i` :man_shrugging: 

maybe that rule I said is not so important for list comprehensions and loop indices. It's fine with me to change it here.

pybv/io.py
```diff
@@ -54,12 +54,13 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
...
-        factor, and all decimals will be cut off after this. So, this number
-        controls the amount of round-trip resolution you want.
-        This can be either a single float for all channels or an array with
-        nchan elements.
+        By default, this will be 1e-7, or 0.1 µV. This number controls the
+        amount of round-trip resolution. This can be either a single float for
+        all channels or an array with n_channels elements.
+    unit : str
+        The unit of the exported data. This can be one of 'V', 'mV', 'µV' (or
+        equivalently 'uV') , 'nV' or 'auto'. If 'auto', a suitable unit based
```
tstenner: I'd prefer `None` instead of `'auto'`

pybv/tests/test_bv_writer.py
tstenner: `grep`ing for variable names is asking for trouble and AFAIK even VSCode without any plugins properly highlights single letter variables. OTOH, if you're looking for expressive variable names and `i` isn't good enough, `index` is better than `ii`.

pybv/tests/test_bv_writer.py
tstenner: I think it's the cartesian product so the test will be run 7*5=35 times

cbrnr: Yep.

pybv/tests/test_bv_writer.py
cbrnr: I agree with @tstenner, in some cases it's fine to use one-letter names, and in other cases more expressive names are needed (and `ii` is not more expressive than `i`).

pybv/io.py
cbrnr: Will do!

pybv/tests/test_bv_writer.py
sappelhoff: thanks for the explanations :+1:

cbrnr:
> cool, let's wait for the MNE merge. Should be soon. Can you please update What's New under the features section?

Done, but the test will fail until there is a new MNE release on PyPI with the change in https://github.com/mne-tools/mne-python/pull/7539 - do we want that?

sappelhoff:
> Done, but the test will fail until there is a new MNE release on PyPI with the change in mne-tools/mne-python#7539 - do we want that?

Options:

1. we can adjust the tests to run `master` of MNE until the next release.
1. we can skip the failing test for now and until the next release of MNE (explicitly marking that test, because we know what's going on and why it's failing)
1. we can wait with this PR until MNE makes a backport of your fix

I prefer option 1. ... and then switching our CIs back to *not* running on MNE `master` once your fix is backported to `0.20`

cbrnr:
I'm OK with 1 - can you make the change in a separate PR and then I rebase?

sappelhoff:
see https://github.com/bids-standard/pybv/commit/dadee0072cffd001be60e5ff9b67358b2811f900

cbrnr:
@sappelhoff could you make a new release? This is an important fix that makes exported files (more) compatible with BrainVision Analyzer.

sappelhoff:
sure, will do today :+1:

sappelhoff:
Done, see:

- https://pypi.org/project/pybv/
- https://github.com/bids-standard/pybv/releases/tag/v0.3.0

cbrnr:
Thanks @sappelhoff!

PR #41, opened 2020-03-30
MAINT: texts, and reorganization

sappelhoff:
I used the time during one of the many mass-phone conferences these days to make minor doc improvements.


sappelhoff:
And we get an unrelated error here that can be fixed elsewhere, I opened #42 for this.

PR #47, opened 2020-11-05
Auto-convert "greek small letter mu" to micro symbol

### LINKED ISSUE #46: "Units" parameter is not checked for valid units.
Passing "μV" fails, because μ is https://www.compart.com/de/unicode/U+03BC, but only µ (https://www.compart.com/de/unicode/U+00B5) is supported in BV according to the spec.

Also passing nonsense like "sV" results in an ungraceful error.

image

sappelhoff:
closes #46

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -5,6 +5,12 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  create:
+    branches: [master]
+    tags: ['**']
+  schedule:
+    - cron: "0 4 1 * *"
+
 
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


.github/workflows/python_tests.yml
```diff
@@ -5,6 +5,12 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  create:
+    branches: [master]
+    tags: ['**']
+  schedule:
+    - cron: "0 4 1 * *"
```
sappelhoff, line 12: runs the CI as a cron job every 1st day of a month at 4am in the night

hoechenberger: Why not every day though?

sappelhoff: I thought cause of the environment :grinning: , and we only need to cover the case where mne or numpy change something - that's not so frequent

sappelhoff: are you happy with this PR otherwise?

hoechenberger (approved):

sappelhoff:
Thanks for the review @hoechenberger!

PR #53, opened 2020-11-06
FIX: support for for commas in channel names

### LINKED ISSUE #51: Currently no support for commas in channel names, event types, and event descriptions
Looking at the first line of this code snippet, we see that "commas in channel names are coded as \1"

https://github.com/bids-standard/pybv/blob/76e8bea84d6594ca8e87f8fe367896167e43c641/pybv/io.py#L269-L281

However, we do not perform this coding ... we simply take a list of channels and print them to the vhdr as they are.

This can lead to bugs and should be fixed.

sappelhoff:
closes #51 

commas in (1) event type, and (2) event description are currently irrelevant, because:

1. type is always "Stimulus" (no commas possible)
2. description is always numeric (no commas possible)

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -292,9 +292,10 @@ def _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq, ch_names,
         units = [unit] * nchan
 
         for i in range(nchan):
+            _ch_name = ch_names[i].replace(',', r'\1')
             resolution, unit = _optimize_channel_unit(resolutions[i], units[i])
             s = r'Ch{}={},,{:0.{precision}f},{}'
-            print(s.format(i + 1, ch_names[i], resolution, unit,
+            print(s.format(i + 1, _ch_name, resolution, unit,
                            precision=max(0, int(np.log10(1 / resolution)))),
                   file=fout)
         print(r'', file=fout)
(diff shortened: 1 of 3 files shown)


hoechenberger (commented):
Test looks good, if you get this one to pass, we should be fine 👍

sappelhoff:
This seems like an MNE-Python reading error to me:

image


as if MNE-Python does not properly read channel names with commas.

pybv/io.py
```diff
@@ -292,9 +292,10 @@ def _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq, ch_names,
         units = [unit] * nchan
 
         for i in range(nchan):
+            _ch_name = ch_names[i].replace(',', r'\1')
             resolution, unit = _optimize_channel_unit(resolutions[i], units[i])
             s = r'Ch{}={},,{:0.{precision}f},{}'
-            print(s.format(i + 1, ch_names[i], resolution, unit,
+            print(s.format(i + 1, _ch_name, resolution, unit,
                            precision=max(0, int(np.log10(1 / resolution)))),
                   file=fout)
```
hoechenberger, line 300: You sure you don't want to switch to f-strings here? ;) This looks painful!!

sappelhoff: I absolutely do ... but let's clean out the bugs first :-D

then fstrings, pathlib instead of os.path, and pytest-tmpdir fixtures instead of mkdtmp

sappelhoff:
this one is blocked until https://github.com/mne-tools/mne-python/pull/8492 is resolved.

sappelhoff:
This passes on this mne branch: https://github.com/mne-tools/mne-python/pull/8492

:slightly_smiling_face: so once that is merged, we can merge here as well

sappelhoff:
merging now to continue development without merge conflicts, and because I am confident in the behavior

PR #56, opened 2020-11-06
miscellaneous maintenance enhancements

sappelhoff:
while going through the codebase I adjusted a few things.

I'll comment in the code about my reasoning.

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -156,7 +156,7 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
 
 
 def _chk_fmt(fmt):
-    """Check that the format string is valid, return BVEF / numpy datatypes."""
+    """Check that the format string is valid, return (BV, numpy) datatypes."""
     if fmt not in SUPPORTED_FORMATS:
         errmsg = ('Data format {} not supported.'.format(fmt) +
                   'Currently supported formats are: ' +
@@ -167,7 +167,6 @@ def _chk_fmt(fmt):
 
 def _chk_multiplexed(orientation):
     """Validate an orientation, return if it is multiplexed or not."""
-    orientation = orientation.lower()
     if orientation not in SUPPORTED_ORIENTS:
         errmsg = ('Orientation {} not supported.'.format(orientation) +
                   'Currently supported orientations are: ' +
@@ -246,28 +245,26 @@ def _optimize_channel_unit(resolution, unit):
 
 
 def _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq, ch_names,
-                     orientation='multiplexed', format='binary_float32',
-                     resolution=1e-7, unit='µV'):
+                     orientation, format, resolution, unit):
     """Write BrainvVision header file."""
-    fmt = format.lower()
     bvfmt, _ = _chk_fmt(format)
 
     multiplexed = _chk_multiplexed(orientation)
 
     with codecs.open(vhdr_fname, 'w', encoding='utf-8') as fout:
         print(r'Brain Vision Data Exchange Header File Version 1.0', file=fout)
-        print(r';Written using pybv {}'.format(__version__), file=fout)
+        print(r'; Written using pybv {}'.format(__version__), file=fout)
         print(r'', file=fout)
         print(r'[Common Infos]', file=fout)
         print(r'Codepage=UTF-8', file=fout)
         print(r'DataFile={}'.format(op.basename(eeg_fname)), file=fout)
         print(r'MarkerFile={}'.format(op.basename(vmrk_fname)), file=fout)
 
-        if fmt.startswith('binary'):
+        if format.startswith('binary'):
             print(r'DataFormat=BINARY', file=fout)
 
         if multiplexed:
-            print(r'; DataOrientation: MULTIPLEXED=ch1,pt1, ch2,pt1 ...', file=fout)  # noqa: E501
+            print(r'; Data orientation: MULTIPLEXED=ch1,pt1, ch2,pt1 ...', file=fout)  # noqa: E501
             print(r'DataOrientation=MULTIPLEXED', file=fout)
 
         print(r'NumberOfChannels={}'.format(len(data)), file=fout)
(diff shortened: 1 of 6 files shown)


pybv/io.py
```diff
@@ -156,7 +156,7 @@ def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
 
 
 def _chk_fmt(fmt):
-    """Check that the format string is valid, return BVEF / numpy datatypes."""
+    """Check that the format string is valid, return (BV, numpy) datatypes."""
```
sappelhoff, line 159: BVEF is BrainVision Electrode File, so this is a typo

pybv/io.py
```diff
@@ -246,43 +245,41 @@ def _optimize_channel_unit(resolution, unit):
 
 
 def _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq, ch_names,
-                     orientation='multiplexed', format='binary_float32',
-                     resolution=1e-7, unit='µV'):
+                     orientation, format, resolution, unit):
```
sappelhoff, line 248: I find it scary to have many default values for private functions with many parameters --> this can easily lead to uncaught errors.

So I remove all defaults except for the public write_raw_brainvision

pybv/io.py
```diff
@@ -246,43 +245,41 @@ def _optimize_channel_unit(resolution, unit):
...
     """Write BrainvVision header file."""
-    fmt = format.lower()
     bvfmt, _ = _chk_fmt(format)
 
     multiplexed = _chk_multiplexed(orientation)
 
     with codecs.open(vhdr_fname, 'w', encoding='utf-8') as fout:
         print(r'Brain Vision Data Exchange Header File Version 1.0', file=fout)
-        print(r';Written using pybv {}'.format(__version__), file=fout)
+        print(r'; Written using pybv {}'.format(__version__), file=fout)
```
sappelhoff, line 256: in the "real" bv files, there is always a space after the start of a comment - so do it here as well

pybv/io.py
```diff
@@ -246,43 +245,41 @@ def _optimize_channel_unit(resolution, unit):
 
 
 def _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq, ch_names,
-                     orientation='multiplexed', format='binary_float32',
-                     resolution=1e-7, unit='µV'):
+                     orientation, format, resolution, unit):
     """Write BrainvVision header file."""
-    fmt = format.lower()
```
sappelhoff, line 252: there is no need to call .lower, because we are checking the validity of the format parameter value and only continue if it's valid.

pybv/io.py
```diff
@@ -246,43 +245,41 @@ def _optimize_channel_unit(resolution, unit):
...
-        print(r';Written using pybv {}'.format(__version__), file=fout)
+        print(r'; Written using pybv {}'.format(__version__), file=fout)
         print(r'', file=fout)
         print(r'[Common Infos]', file=fout)
         print(r'Codepage=UTF-8', file=fout)
         print(r'DataFile={}'.format(op.basename(eeg_fname)), file=fout)
         print(r'MarkerFile={}'.format(op.basename(vmrk_fname)), file=fout)
 
-        if fmt.startswith('binary'):
+        if format.startswith('binary'):
```
sappelhoff, line 263: see above, I removed `fmt = format.lower()` because it was unnecessary

pybv/io.py
```diff
@@ -246,43 +245,41 @@ def _optimize_channel_unit(resolution, unit):
...
             print(r'DataFormat=BINARY', file=fout)
 
         if multiplexed:
-            print(r'; DataOrientation: MULTIPLEXED=ch1,pt1, ch2,pt1 ...', file=fout)  # noqa: E501
+            print(r'; Data orientation: MULTIPLEXED=ch1,pt1, ch2,pt1 ...', file=fout)  # noqa: E501
```
sappelhoff, line 267: this is how it's written in the "real" BV files. ... may not make sense, but let's conform

pybv/io.py
```diff
@@ -246,43 +245,41 @@ def _optimize_channel_unit(resolution, unit):
...
-        if fmt.startswith('binary'):
+        if format.startswith('binary'):
             print(r'[Binary Infos]', file=fout)
             print(r'BinaryFormat={}'.format(bvfmt), file=fout)
             print(r'', file=fout)
 
         print(r'[Channel Infos]', file=fout)
         print(r'; Each entry: Ch<Channel number>=<Name>,<Reference channel name>,', file=fout)  # noqa: E501
-        print(r'; <Resolution in "unit">,<unit>,Future extensions…', file=fout)
+        print(r'; <Resolution in "Unit">,<Unit>, Future extensions..', file=fout)  # noqa: E501
```
sappelhoff, line 282: again, copying how it is in "real" BV files --> including the oddly two (!) trailing dots

pybv/tests/test_bv_writer.py
```diff
@@ -6,18 +6,15 @@
...
 from shutil import rmtree
 from tempfile import mkdtemp
-from datetime import datetime, timezone
-
-import pytest
 
 import mne
 import numpy as np
 from numpy.testing import assert_allclose, assert_array_equal
+import pytest
```
sappelhoff, line 17: ordering imports here, maybe we should add isort to the CI

pybv/tests/test_bv_writer.py
```diff
@@ -154,7 +152,7 @@ def test_write_read_cycle(meas_date):
 
     assert_allclose(data, raw_written._data)  # data round-trip
 
-    assert ch_names == raw_written.ch_names  # channels
+    assert_array_equal(ch_names, raw_written.ch_names)  # channels
```
sappelhoff, line 155: assert_array_equal provides a much more detailed diff in case of a failure

pybv/tests/test_bv_writer.py
```diff
@@ -178,8 +176,15 @@ def test_unit_resolution(resolution, unit):
...
+    vhdr_fname = os.path.join(tmpdir, fname + '.vhdr')
     raw_written = mne.io.read_raw_brainvision(vhdr_fname, preload=True)
-    assert np.allclose(data, raw_written.get_data())
+    assert_allclose(data, raw_written.get_data())
+
+    # Check that the correct units were written in the BV file
+    orig_units = [u for key, u in raw_written._orig_units.items()]
+    assert len(set(orig_units)) == 1
+    if unit is not None:
+        assert orig_units[0] == unit.replace("u", "µ")
```
sappelhoff, line 188: small double/sanity check

pybv/tests/test_bv_writer.py
sappelhoff: I did add isort and added it to the CI --> doing that early for a project leads to much more readable import sections in the long run.

pybv/io.py
hoechenberger: In addition, switching to keyword-only arguments can make total sense for functions with that many parameters!

hoechenberger (approved):

pybv/io.py
sappelhoff: will do in a follow up pr

PR #57, opened 2020-11-07
API: Make write_brainvision take keyword arguments only

sappelhoff:
`write_brainvision` is a long function with many parameters.

Making these parameters keyword only gives us two advantages:

- during development we can in the future move around the parameters as much as we want without breaking anybody's code
- it forces the user to be aware of what argument they pass to which parameter, and errors are less likely

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -30,7 +30,7 @@ SUPPORTED_ORIENTS = {'multiplexed'}
 SUPPORTED_UNITS = ['V', 'mV', 'µV', 'uV', 'nV']
 
 
-def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
+def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
                       events=None, resolution=1e-7, unit='µV', scale_data=True,
                       fmt='binary_float32', meas_date=None):
     """Write raw data to BrainVision format.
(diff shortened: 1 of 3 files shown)


pybv/io.py
```diff
@@ -30,7 +30,7 @@
 SUPPORTED_UNITS = ['V', 'mV', 'µV', 'uV', 'nV']
 
 
-def write_brainvision(data, sfreq, ch_names, fname_base, folder_out,
+def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
```
hoechenberger, line 33: I guess I would allow for `data` to be passed by position, but it's a matter of taste I suppose!
```suggestion
def write_brainvision(data, *, sfreq, ch_names, fname_base, folder_out,
```

sappelhoff: I would like to keep it all keyword, no positional because I think it leads to a better consistency in this case

cbrnr: I'd also prefer the first argument to not be kw-only, but this writer is different enough that I'm also fine with all kw-only (normally the data is the first and the file name the second parameter).

hoechenberger: I'm fine with either, just meant to throw in another possibility :)

sappelhoff: Okay great, thanks both! I am merging then and following up with a huge API PR that'll better align what we are doing with the BrainVision spec.

Issue #49, opened 2020-11-05
Wrong resolution in absolute terms is printed to vhdr


sappelhoff:
this is a separate issue from #48

looking at the following code, we see that the "resolution" for each channel first passes through "optimize_channel_unit"

https://github.com/bids-standard/pybv/blob/76e8bea84d6594ca8e87f8fe367896167e43c641/pybv/io.py#L276-L281

Yet "optimize_channel_unit" changes the resolution to account for conversion of units:

https://github.com/bids-standard/pybv/blob/76e8bea84d6594ca8e87f8fe367896167e43c641/pybv/io.py#L210-L227

This is the desired behavior for scaling the data at the same time to both (1) from volts to the desired unit, and (2) to the desired resolution. HOWEVER, in the VHDR, the unit and the resolution are noted for each channel. Thus the original resolution *prior* to going through "optimize_channel_unit" must be written to VHDR

sappelhoff:
@tstenner do you happen to have time to look into this? :slightly_smiling_face:

sappelhoff:
we should also add a test that does *not* depend on MNE.

-->

1. write data to bv given a vector of different resolutions
1. read the resulting vhdr file with bare python
1. check that the written and original resolutions match exactly

sappelhoff:
actually, I am seeing now that we never scale the data to the desired unit. The data is only scaled to the desired resolution in VOLTS, and then written to file.

Then the "resolution" field in vhdr is "abused" to encode both the conversion from volt to the desired unit AND the back-scaling.

This works equivalently, but it's not what the BrainVision spec says. --> the resolution is supposed to be the resolution *in units*.

Looking at the docstr for resolution I see, that we also require that to be "in volts", instead of in "units" --> so at least we are being consistent there, and it explains why the float32 format roundtrips all pass.

I still think we should change this behavior to be more in line with the BV standard.

And I still have no solution for why the int16 tests fail. #45

PR #58, opened 2020-11-07
API, HUGE: rework "resolution" and follow the consequences

### LINKED ISSUE #48: "resolution" is printed with wrong number of decimals if not in [..., 1e-6, 1e-5, 1e-4, ...]
This is how resolution is saved to the vhdr file:

https://github.com/bids-standard/pybv/blob/76e8bea84d6594ca8e87f8fe367896167e43c641/pybv/io.py#L276-L280

consider the way "precision" currently determines how "resolution" is printed (the decimal points).

It works well for 1e-6, 1e-7, 1e-8, ...

```python
res = 1.e-3
mystr = "{:0.{precision}f}".format(res, precision=max(0, int(np.log10(1/res)))) 
print(mystr)

# 0.001
```

but it fails for e.g., 1.1e-3

```python
res = 1.1e-3
mystr = "{:0.{precision}f}".format(res, precision=max(0, int(np.log10(1/res)))) 
print(mystr)

# 0.00
```

sappelhoff:
closes https://github.com/bids-standard/pybv/issues/49

also closes closes #48

this commit makes some huge changes in order to align the way pybv
saves data more with the BrainVision data format specification.

- the data is now converted to units, and then scaled to resolution,
  and only then written to disk

- resolution is written *as is* to the VHDR file (as is the case for
  unit)

- passing None for unit is removed, unit has to be passed explicitly
  (defaults to µV)

- the scale_data parameter is removed entirely, because with the new
  way that resolution works, it has become meaningless and passing
  resolution=1 takes over the same functionality

to do:

- [x] add whatsnew
- ~rename unit_resolution test to unit_resolution_format or the like~ will do in follow-up PR

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -34,6 +34,8 @@ API
 
 Authors
 ~~~~~~~
+- `Clemens Brunner`_
+- `Richard Höchenberger`_
 - `Stefan Appelhoff`_
 
 0.3.0
@@ -117,3 +119,4 @@ Authors
 .. _Tristan Stenner: https://github.com/tstenner
 .. _Phillip Alday: https://palday.bitbucket.io/
 .. _Clemens Brunner: https://cbrnr.github.io/
+.. _Richard Höchenberger: https://hoechenberger.net/
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,16 +17,19 @@ Current
 
 Changelog
 ~~~~~~~~~
-- Passing a "greek small letter mu" instead of a "micro sign" as a ``unit`` is now permitted, because :func:`pybv.write_brainvision` will convert from one to the other, by `Stefan Appelhoff`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
+- Passing a "greek small letter mu" to the ``unit`` parameter in :func:`pybv.write_brainvision` instead of a "micro sign" is now permitted, because the one will be automatically convert to the other, by `Stefan Appelhoff`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
 
 Bug
 ~~~
-- ``pybv`` now properly substitutes commas in channel names with ``\1`` before writing to file, and passing non-numeric events now raises a ValueError, by `Stefan Appelhoff`_ (`#53 <https://github.com/bids-standard/pybv/pull/53>`_)
+- Fix bug where :func:`pybv.write_brainvision` did not properly deal with commas in channel names and non-numeric events, by `Stefan Appelhoff`_ (`#53 <https://github.com/bids-standard/pybv/pull/53>`_)
 - :func:`pybv.write_brainvision` now properly handles sampling frequencies that are not multiples of 10 (even floats), by `Clemens Brunner`_ (`#59 <https://github.com/bids-standard/pybv/pull/59>`_)
+- Fix bug where :func:`pybv.write_brainvision` would write a different resolution to the ``vhdr`` file than specified with the ``resolution`` parameter. Note that this did *not* affect the roundtrip accuracy of the written data, because of internal scaling of the data, by `Stefan Appelhoff`_ (`#58 <https://github.com/bids-standard/pybv/pull/58>`_)
 
 API
 ~~~
 - :func:`pybv.write_brainvision` now accepts keyword arguments only. Positional arguments are no longer allowed, by `Stefan Appelhoff`_ (`#57 <https://github.com/bids-standard/pybv/pull/57>`_)
+- In :func:`pybv.write_brainvision`, the ``scale_data`` parameter was removed from :func:`pybv.write_brainvision`, by `Stefan Appelhoff`_ (`#58 <https://github.com/bids-standard/pybv/pull/58>`_)
+- In :func:`pybv.write_brainvision`, the ``unit`` parameter no longer accepts an argument ``None`` to automatically determined a unit based on the ``resolution``, by `Stefan Appelhoff`_ (`#58 <https://github.com/bids-standard/pybv/pull/58>`_)
 
 Authors
 ~~~~~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,6 +24,7 @@ Bug
 - Fix bug where :func:`pybv.write_brainvision` did not properly deal with commas in channel names and non-numeric events, by `Stefan Appelhoff`_ (`#53 <https://github.com/bids-standard/pybv/pull/53>`_)
 - :func:`pybv.write_brainvision` now properly handles sampling frequencies that are not multiples of 10 (even floats), by `Clemens Brunner`_ (`#59 <https://github.com/bids-standard/pybv/pull/59>`_)
 - Fix bug where :func:`pybv.write_brainvision` would write a different resolution to the ``vhdr`` file than specified with the ``resolution`` parameter. Note that this did *not* affect the roundtrip accuracy of the written data, because of internal scaling of the data, by `Stefan Appelhoff`_ (`#58 <https://github.com/bids-standard/pybv/pull/58>`_)
+- Fix bug where values for the ``resolution`` parameter like ``0.5``, ``0.123``, ``3.143`` were not written with adequate decimal precision in :func:`pybv.write_brainvision`, by `Stefan Appelhoff`_ (`#58 <https://github.com/bids-standard/pybv/pull/58>`_)
 
 API
 ~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,7 +17,7 @@ Current
 
 Changelog
 ~~~~~~~~~
-- Passing a "greek small letter mu" to the ``unit`` parameter in :func:`pybv.write_brainvision` instead of a "micro sign" is now permitted, because the one will be automatically convert to the other, by `Stefan Appelhoff`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
+- Passing a "greek small letter mu" to the ``unit`` parameter in :func:`pybv.write_brainvision` instead of a "micro sign" is now permitted, because the former will be automatically convert to the latter, by `Stefan Appelhoff`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
 
 Bug
 ~~~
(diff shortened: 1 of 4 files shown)


pybv/io.py
```diff
@@ -59,20 +59,18 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
-        amount of round-trip resolution. This can be either a single float for
-        all channels or an array with n_channels elements.
-    unit : str | None
+    resolution : float | ndarray, shape(nchannels,)
+        The resolution in `unit` in which you'd like the data to be stored.
+        If float, the same resolution is applied to all channels.
+        If ndarray with n_channels elements, each channel is scaled with
+        its own corresponding resolution from the ndarray.
+        Note that if `fmt` is 'binary_int16', it is recommended to set
+        `resolution` to '0.1' (default) and `unit` to 'µV' (default).
```
sappelhoff, line 68: actually, "binary_int16" shouldn't be used at all until we figure out why roundtrips are failing

sappelhoff:
Note how the minimal (and unrelated) changes to the tests are proof that the two ways to deal with "resolution" are equivalent ... but the way proposed in this PR is what the BrainVision spec *actually* says should be done.

@hoechenberger @cbrnr

pybv/io.py
```diff
@@ -31,14 +31,14 @@
 
 
 def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
-                      events=None, resolution=1e-7, unit='µV', scale_data=True,
+                      events=None, resolution=0.1, unit='µV',
```
hoechenberger, line 34: So this means we're ensuring a correct roundtrip for values >= 0.1 µV by default, yes? Could this be made a little bit more explicit when describing the `resolution` param?

pybv/io.py
```diff
@@ -226,30 +224,20 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date):
...
     elif unit == 'nV':
-        return resolution / 1e-9, 'nV'
+        return data * 1e9
     else:
         raise ValueError(
             f'Encountered unsupported unit: {unit}'
-            '\nUse either "None" for `unit`, or one of the following: '
-            f'{SUPPORTED_UNITS}'
-            )
+            '\nUse one of the following: {SUPPORTED_UNITS}')
```
hoechenberger, line 240: I believe you'd need to add an `f` here too; and even if it's optional, I'd still add it for visual consistency with the previous line :)

pybv/io.py
sappelhoff: I actually don't get the resolution parameter beyond the fact that before saving we do "data *= (1/resolution)" and at reading, readers do "data *= resolution". I don't get what that buys us
... Can you explain that to me?

sappelhoff: Ohhh because we are writing to int! So when writing to float, we can completely ignore resolution!

sappelhoff:
I think I know why int16 roundtrip is failing ... sometimes we scale the data so high, that it cannot be represented in int16 anymore ... whereas it's totally possible in float32.

We need to add a check that makes use of `np.iinfo` and `np.finfo`:

```Python
In [20]: np.iinfo(np.int16)                                                                                                    
Out[20]: iinfo(min=-32768, max=32767, dtype=int16)

In [21]: np.finfo(np.float32)                                                                                                  
Out[21]: finfo(resolution=1e-06, min=-3.4028235e+38, max=3.4028235e+38, dtype=float32)

```

If we detect that after scaling the data, it exceeds thresholds, we raise an error to tell the user to use a different unit and/or resolution, making a concrete proposal that would be in the limits of int16.

same for float32, but there is much more room there until we'd hit the max (or min) of what can be represented :-) 

(also good to know that default resolution for float32 is 1e-6)

pybv/io.py
sappelhoff: I figured it out now, will push a follow up PR after this one.

pybv/io.py
sappelhoff: agreed, done!

cbrnr: Some linters will complain if they find an f-string without any curly braces; if that's the case be sure to `noqa` these lines. I agree that for continued strings it looks better to let all lines be f-strings.

pybv/io.py
```diff
@@ -64,10 +64,12 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
         If ndarray with n_channels elements, each channel is scaled with
         its own corresponding resolution from the ndarray.
-        Note that if `fmt` is 'binary_int16', it is recommended to set
-        `resolution` to '0.1' (default) and `unit` to 'µV' (default).
-        For `fmt` 'binary_float32', the resolution should usually be kept
-        at its default value.
+        Note that `resolution` is applied on top of the default resolution
+        that a data format (see `fmt`) has; that is, 0 floating point
+        resolution for binary_int16, and 1e-6 floating point resolution for
+        binary_float32. For example, writing binary_int16 data in µV with
```
cbrnr, line 70: This is still not clear to me. What does "0 floating point resolution for binary_int16" mean?

sappelhoff: What I want to express with this, is that if you write, say,  `1.234` to int16, you will end up with `1`, the floating point resolution is 0. Does that make sense?

docs/changelog.rst
```diff
@@ -17,15 +17,18 @@ Current
 
 Changelog
 ~~~~~~~~~
-- Passing a "greek small letter mu" instead of a "micro sign" as a ``unit`` is now permitted, because :func:`pybv.write_brainvision` will convert from one to the other, by `Stefan Appelhoff`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
+- Passing a "greek small letter mu" to the ``unit`` parameter in :func:`pybv.write_brainvision` instead of a "micro sign" is now permitted, because the one will be automatically convert to the other, by `Stefan Appelhoff`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
```
cbrnr, line 20:
```suggestion
- Passing a "greek small letter mu" to the ``unit`` parameter in :func:`pybv.write_brainvision` instead of a "micro sign" is now permitted, because the former will be automatically convert to the latter, by `Stefan Appelhoff`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
```

docs/changelog.rst
```diff
@@ -17,15 +17,18 @@ Current
...
+- Fix bug where :func:`pybv.write_brainvision` did not properly deal with commas in channel names and non-numeric events, by `Stefan Appelhoff`_ (`#53 <https://github.com/bids-standard/pybv/pull/53>`_)
+- Fix bug where :func:`pybv.write_brainvision` would write a different resolution to the ``vhdr`` file than specified with the ``resolution`` parameter. Note that this did *not* affect the roundtrip accuracy of the written data, because of internal scaling of the data, by `Stefan Appelhoff`_ (`#58 <https://github.com/bids-standard/pybv/pull/58>`_)
 
 API
 ~~~
 - :func:`pybv.write_brainvision` now accepts keyword arguments only. Positional arguments are no longer allowed, by `Stefan Appelhoff`_ (`#57 <https://github.com/bids-standard/pybv/pull/57>`_)
+- In :func:`pybv.write_brainvision`, the ``scale_data`` parameter was removed from :func:`pybv.write_brainvision`, by `Stefan Appelhoff`_ (`#58 <https://github.com/bids-standard/pybv/pull/58>`_)
+- In :func:`pybv.write_brainvision`, the ``unit`` parameter no longer accepts an argument ``None`` to automatically determined a unit based on the ``resolution``, by `Stefan Appelhoff`_ (`#58 <https://github.com/bids-standard/pybv/pull/58>`_)
```
cbrnr, line 31:
```suggestion
- In :func:`pybv.write_brainvision`, the ``unit`` parameter no longer accepts an argument ``None`` to automatically determine a unit based on the ``resolution``, by `Stefan Appelhoff`_ (`#58 <https://github.com/bids-standard/pybv/pull/58>`_)
```

pybv/io.py
sappelhoff: @cbrnr @hoechenberger for this param docstr your input would be highly valued, so that we can make it straight forward for users

cbrnr: Ah I see! I think it is confusing to talk about a "floating point resolution" when you want to save an int. I think the example already makes it pretty clear, so what about deleting the sentence starting with "Note that..."?

sappelhoff: I think that would be okay, and less confusing --> but it also removes some information. For example `np.finfo(np.float32)` tells me that the resolution of float32 on my system is `1e-6` --> now if I apply the pybv resolution of `0.1`, I'll actually have a `1e-7` resolution, not a `0.1` resolution - or am I overthinking this?

cbrnr: Hm. This is really pretty complicated. The resolution of float32 will always be 1e-6, but the resolution parameter effectively scales the data so you can shift the unit and therefore effectively change the resolution. I really wonder why they decided to include such a parameter, it's probably only necessary if you store integer data. Maybe we should focus on this use case in our explanations?

sappelhoff:
I'll leave this open tomorrow as well in the hopes for new suggestions on how to get a better docstring.

But in the meantime, this is the PR that'll fix the int_16 issue. I based it on *this* PR and will open it to pybv once *this* PR is merged: https://github.com/sappelhoff/pybv/pull/1

sappelhoff (commented):
I updated the docstr param. It now reads like this:

```
    resolution : float | ndarray, shape(nchannels,)
        The resolution in `unit` in which you'd like the data to be stored.
        If float, the same resolution is applied to all channels.
        If ndarray with n_channels elements, each channel is scaled with
        its own corresponding resolution from the ndarray.
        Note that `resolution` is applied on top of the default resolution
        that a data format (see `fmt`) has. For example, the binary_int16
        format by design has no floating point support, but when scaling the
        data in µV for 0.1 resolution (default), accurate writing for all
        values >= 0.1 µV will be guaranteed. In contrast, the binary_float32
        format by design already supports floating points up to 1e-6
        resolution, and writing data in µV with 0.1 resolution will thus
        guarantee accurate writing vor all values >= 1e-7 µV
        (``1e-6 * 0.1``).
```

sappelhoff:
moving on, thanks for reviews

PR #59, opened 2020-11-07
Sampling interval can be float

### LINKED ISSUE #52: float sampling frequencies are not supported
the BrainVision format specifies the sampling frequency of the data as "sampling interval in microseconds"

We are converting the sampling frequency in Hertz like this:

https://github.com/bids-standard/pybv/blob/76e8bea84d6594ca8e87f8fe367896167e43c641/pybv/io.py#L257

This means that we do not currently support sampling frequencies like `11.5`  ... I am not even sure whether that can reasonably exists, but we should add a check for that in the code anyways.

cbrnr:
Fixes #52. I've also introduced f-strings in this block of code - do you want me to convert the rest as well?

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -22,6 +22,7 @@ Changelog
 Bug
 ~~~
 - ``pybv`` now properly substitutes commas in channel names with ``\1`` before writing to file, and passing non-numeric events now raises a ValueError, by `Stefan Appelhoff`_ (`#53 <https://github.com/bids-standard/pybv/pull/53>`_)
+- :func:`pybv.write_brainvision` now properly handles sampling frequencies (even floats), by `Clemens Brunner`_ (`#59 <https://github.com/bids-standard/pybv/pull/59>`_)
 
 API
 ~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -22,7 +22,7 @@ Changelog
 Bug
 ~~~
 - ``pybv`` now properly substitutes commas in channel names with ``\1`` before writing to file, and passing non-numeric events now raises a ValueError, by `Stefan Appelhoff`_ (`#53 <https://github.com/bids-standard/pybv/pull/53>`_)
-- :func:`pybv.write_brainvision` now properly handles sampling frequencies (even floats), by `Clemens Brunner`_ (`#59 <https://github.com/bids-standard/pybv/pull/59>`_)
+- :func:`pybv.write_brainvision` now properly handles sampling frequencies that are not multiples of 10 (even floats), by `Clemens Brunner`_ (`#59 <https://github.com/bids-standard/pybv/pull/59>`_)
 
 API
 ~~~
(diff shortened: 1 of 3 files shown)


sappelhoff (commented):
Cool, thanks.

Can you please add a test covering a bunch of conventionally used sfreqs? (e.g. 512, 125, 100, 500, 1000, 1024, ...)

re: f-strings yes, would be nice! Preferably as a follow up PR though!

sappelhoff:
Finally, please add a what's new entry --> I think this would fall under the "bug" category, WDYT?

cbrnr:
@sappelhoff is this test OK?

pybv/tests/test_bv_writer.py
```diff
@@ -243,3 +243,16 @@ def test_unit_resolution(resolution, unit):
...
+def test_sampling_frequencies(sfreq):
+    """Test different sampling frequencies."""
+    tmpdir = _mktmpdir()
+    # sampling frequency gets written as sampling interval
+    write_brainvision(data=data, sfreq=sfreq, ch_names=ch_names,
+                      fname_base=fname, folder_out=tmpdir, scale_data=False)
+    vhdr_fname = os.path.join(tmpdir, fname + '.vhdr')
+    raw_written = mne.io.read_raw_brainvision(vhdr_fname=vhdr_fname,
+                                              preload=True)
+    assert_allclose(sfreq, raw_written.info['sfreq'])
```
sappelhoff, line 258:
```suggestion
    assert_allclose(sfreq, raw_written.info['sfreq'])
    rmtree(tmpdir)    
```

docs/changelog.rst
```diff
@@ -22,6 +22,7 @@ Changelog
 Bug
 ~~~
 - ``pybv`` now properly substitutes commas in channel names with ``\1`` before writing to file, and passing non-numeric events now raises a ValueError, by `Stefan Appelhoff`_ (`#53 <https://github.com/bids-standard/pybv/pull/53>`_)
+- :func:`pybv.write_brainvision` now properly handles sampling frequencies (even floats), by `Clemens Brunner`_ (`#59 <https://github.com/bids-standard/pybv/pull/59>`_)
```
sappelhoff, line 25:
```suggestion
- :func:`pybv.write_brainvision` now properly handles sampling frequencies that are no multiples of 10 (even floats), by `Clemens Brunner`_ (`#59 <https://github.com/bids-standard/pybv/pull/59>`_)
```

sappelhoff (approved):
LGTM apart from my two suggestions.

If tests pass, this can be merged!

cbrnr:
But do you think it's necessary to read the interval from the VHDR file and compare that with the inverse of the specified sampling frequency? Or do we just rely on MNE to do that during loading?

sappelhoff:
> But do you think it's necessary to read the interval from the VHDR file and compare that with the inverse of the specified sampling frequency? Or do we just rely on MNE to do that during loading?

I think ideally, we'd do both :thinking:  but for now MNE is probably sufficient.

sappelhoff:
Thanks @cbrnr if you are up to add a test with bare-python reading+checking the written sfreq, that'd be a nice addon :)

PR #61, opened 2020-11-07
Use f-strings

### LINKED ISSUE #54: switch to f-strings
We now require Python 3.6 for pybv and we can thus switch to using fstrings like: `print(f"this nice value: {value}")` instead of `print("this nice value: {}".format(value))`

@hoechenberger and I think that that'll make the codebase more readable.

cbrnr:
Fixes #54. Please double-check if all changes are OK (there were some tricky ones).

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -120,7 +120,7 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
 
     resolution = np.atleast_1d(resolution)
     if not np.issubdtype(resolution.dtype, np.number):
-        raise ValueError("Resolution should be numeric, is {resolution.dtype}")
+        raise ValueError(f"Resolution should be numeric, is {resolution.dtype}")
 
     if resolution.shape != (1,) and resolution.shape != (nchan,):
         raise ValueError("Resolution should be one or n_chan floats")
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -108,13 +108,11 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
     nchan = len(ch_names)
 
     if len(data) != nchan:
-        raise ValueError("Number of channels in data ({}) does "
-                         "not match number of channel names ({})"
-                         .format(len(data), len(ch_names)))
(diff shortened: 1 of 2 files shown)


pybv/io.py
```diff
@@ -108,21 +108,19 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
+        raise ValueError("Channel names must be unique, found duplicate name.")
 
     if not isinstance(sfreq, (int, float)):
         raise ValueError("sfreq must be one of (float | int)")
     sfreq = float(sfreq)
 
     resolution = np.atleast_1d(resolution)
     if not np.issubdtype(resolution.dtype, np.number):
-        raise ValueError("Resolution should be numeric, is {}".format(resolution.dtype))  # noqa: E501
+        raise ValueError("Resolution should be numeric, is {resolution.dtype}")
```
hoechenberger, line 123:
```suggestion
        raise ValueError(f"Resolution should be numeric, is {resolution.dtype}")
```

hoechenberger (approved):

pybv/io.py
```diff
@@ -108,21 +108,19 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
+        raise ValueError("Channel names must be unique, found duplicate name.")
 
     if not isinstance(sfreq, (int, float)):
         raise ValueError("sfreq must be one of (float | int)")
     sfreq = float(sfreq)
 
     resolution = np.atleast_1d(resolution)
     if not np.issubdtype(resolution.dtype, np.number):
-        raise ValueError("Resolution should be numeric, is {}".format(resolution.dtype))  # noqa: E501
+        raise ValueError(f"Resolution should be numeric, is {resolution.dtype}")
```
sappelhoff, line 123:
```suggestion
        raise ValueError(f"Resolution should be numeric, is {resolution.dtype}")  # noqa:E501
```

sappelhoff (approved):
LGTM, thanks.

I assume you removed the "raw" identifier in front of strs (`r"str"`) because it was not really needed?

cbrnr:
> I assume you removed the "raw" identifier in front of strs (`r"str"`) because it was not really needed?

Yes.

PR #62, opened 2020-11-08
FIX: do not write data if it cannot be represented in fmt, given unit and resolution

### LINKED ISSUE #45: Roundtrip fails for data in "short" format when writing to "binary_int16"
first discovered in https://github.com/mne-tools/mne-bids/pull/610

@hoechenberger prepared a MWE to reproduce the issue:

```Python
# %%
from pathlib import Path
import numpy as np
import pybv
import mne

raw_fname = (Path(mne.__file__).parent / 'io' / 'brainvision' / 'tests' /
             'data' / 'test.vhdr')
raw = mne.io.read_raw_brainvision(raw_fname, preload=True)

raw.pick_types(meg=False, eeg=True)  # to ensure we have only Volt data

assert raw.orig_format == 'short'

pybv.write_brainvision(data=raw.get_data(),
                       sfreq=raw.info['sfreq'],
                       ch_names=raw.ch_names,
                       fname_base='test',
                       folder_out='/tmp',
                       fmt='binary_int16')

raw_read = mne.io.read_raw_brainvision('/tmp/test.vhdr', preload=True)
assert np.allclose(raw.get_data(), raw_read.get_data())
```

--> fails.

Assertion holds if using fmt='binary_float32', although it should be totally fine to write short data as binary_int16

sappelhoff:
fixes #45 

previously we allowed users to write their data in whatever combination of fmt, unit, and resolution they wanted. --> in some cases this lead to data not being accurately represented ... for example, high resolution µV data in  binary_int16 format.

In this PR I add a check to see if the data *can* be written as desired ... and if not, I raise a ValueError.

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -232,8 +232,8 @@ def _scale_data_to_unit(data, unit):
     """Scale `data` in Volts to `data` in `unit`."""
     scale = SUPPORTED_UNITS.get(unit, None)
     if not isinstance(scale, float):
-        msg = (f'Encountered unsupported unit: {unit}'
-               f'\nUse one of the following: {set(SUPPORTED_UNITS.keys())}')
+        msg = (f'Encountered unsupported unit: {unit}\n'
+               f'Use one of the following: {set(SUPPORTED_UNITS.keys())}')
         raise ValueError(msg)
 
     return data * scale
@@ -295,7 +295,7 @@ def _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq, ch_names,
 
 
 def _check_data_in_range(data, dtype):
-    """Check that data can be represented in dtype."""
+    """Check that data can be represented by dtype."""
     check_funcs = {np.int16: np.iinfo, np.float32: np.finfo}
     fun = check_funcs.get(dtype, None)
     if fun is None:  # pragma: no cover
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -231,7 +231,7 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date):
 def _scale_data_to_unit(data, unit):
     """Scale `data` in Volts to `data` in `unit`."""
     scale = SUPPORTED_UNITS.get(unit, None)
-    if not isinstance(scale, float):
+    if scale is None:
         msg = (f'Encountered unsupported unit: {unit}\n'
                f'Use one of the following: {set(SUPPORTED_UNITS.keys())}')
         raise ValueError(msg)
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -28,7 +28,7 @@ SUPPORTED_FORMATS = {
 
 SUPPORTED_ORIENTS = {'multiplexed'}
 
-SUPPORTED_UNITS = ['V', 'mV', 'µV', 'uV', 'nV']
+SUPPORTED_UNITS = {'V': 1e0, 'mV': 1e3, 'µV': 1e6, 'uV': 1e6, 'nV': 1e9}
 
 
 def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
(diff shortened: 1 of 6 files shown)


pybv/io.py
```diff
@@ -229,18 +229,13 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date):
...
-    elif unit in ('µV', 'uV'):
-        return data * 1e6
-    elif unit == 'nV':
-        return data * 1e9
-    else:
-        raise ValueError(
-            f'Encountered unsupported unit: {unit}'
-            f'\nUse one of the following: {SUPPORTED_UNITS}')
+    scale = SUPPORTED_UNITS.get(unit, None)
+    if not isinstance(scale, float):
```
cbrnr, line 234: Shouldn't `if scale is not None` also work?

pybv/io.py
```diff
@@ -229,18 +229,13 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date):
...
+        msg = (f'Encountered unsupported unit: {unit}'
+               f'\nUse one of the following: {set(SUPPORTED_UNITS.keys())}')
```
cbrnr, line 236: I'd move the `\n` to the end of the previous line.

pybv/io.py
```diff
@@ -298,6 +293,18 @@ def _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq, ch_names,
         print('', file=fout)
 
 
+def _check_data_in_range(data, dtype):
+    """Check that data can be represented in dtype."""
```
cbrnr, line 298: in -> by

pybv/tests/test_bv_writer.py
```diff
@@ -218,32 +221,55 @@ def test_write_read_cycle(meas_date):
...
 
 
-@pytest.mark.parametrize("format", ["binary_float32"])
+@pytest.mark.parametrize("format", SUPPORTED_FORMATS.keys())
 @pytest.mark.parametrize("resolution", resolutions)
-@pytest.mark.parametrize("unit", ["V", "mV", "uV", "µV", "nV"])
-def test_unit_resolution(format, resolution, unit):
-    """Test different combinations of units and resolutions."""
+@pytest.mark.parametrize("unit", SUPPORTED_UNITS)
+def test_format_resolution_unit(format, resolution, unit):
```
cbrnr, line 231: It's really hard to see which combination of parameters are expected to fail and which are expected to pass - maybe split the test into several shorter ones?

sappelhoff: I agree that this test is getting out of hand in terms of readability. 

I do not explicitly know beforehand which tests will work and which won't --> the test checks in two ways:

1. data that *can* be represented is written and a roundtrip check is done
1. data that *cannot*, is checked that it raises an error

I would defer improvements of this to future PRs

pybv/io.py
sappelhoff: yes, it should work equivalently, but I found isinstance(..., float) more specific, because we do definitely need a float next :thinking:

pybv/io.py
sappelhoff: agreed, looks prettier. will do

pybv/io.py
cbrnr: But you either get a `float` or `None`, so I find this instance check unnecessary. I'd avoid `isinstance` whenever possible.

pybv/tests/test_bv_writer.py
cbrnr: OK, can you please open an issue as a reminder to simplify this test? Ideally, you'd hardcode a couple of value combinations that work and some others that don't.

pybv/io.py
sappelhoff: I am fine with changing this (will do), but why would you avoid isinstance?

pybv/tests/test_bv_writer.py
sappelhoff: done: https://github.com/bids-standard/pybv/issues/64

pybv/io.py
cbrnr: It isn't Pythonic IMO, because as a dynamically typed language I think code should rely on duck typing as far as possible. It is not always possible, so `isinstance` can't be avoided every time, but in this example a comparison with `None` is completely sufficient.

cbrnr (approved):

pybv/io.py
sappelhoff: thanks for the explanation, makes sense!

Issue #60, opened 2020-11-07
REL 0.4


sappelhoff:
After lots of work during the past few days, I would like to make a 0.4 release that contains lots of bug fixes, API changes, and a more robust and better pybv :slightly_smiling_face: 

There are still many enhancements that we could do (first and foremost #50), but let's put them into 0.5 and following releases.

I think we can do #54 and #55 before the 0.4 release, and of course #58 needs to be merged, and the int16 issue-fix PR opened and merged.

sappelhoff:
done: https://github.com/bids-standard/pybv/releases/tag/v0.4.0

PR #65, opened 2020-11-08
MISC maintenance

### LINKED ISSUE #55: switch "manual" tempdirs for testing to pytest tmpdir fixtures
Making use of this feature: https://docs.pytest.org/en/stable/tmpdir.html

Replacing this code:

https://github.com/bids-standard/pybv/blob/54da125d09b3b3fb83178e7fe06996929aa1ef3d/pybv/tests/test_bv_writer.py#L103-L113

... with the following:

```diff
 @pytest.mark.parametrize("meas_date,match", 
                          [(1, '`meas_date` must be of type str, datetime'), 
                           ('', 'Got a str for `meas_date`, but it was'), 
                           ('1973', 'Got a str for `meas_date`, but it was')]) 
+ def test_bad_meas_date(tmpdir, meas_date, match): 
     """Test that bad measurement dates raise errors.""" 
-     tmpdir = _mktmpdir() 
     with pytest.raises(ValueError, match=match): 
         write_brainvision(data, sfreq, ch_names, fname, tmpdir, 
                           meas_date=meas_date) 
-     rmtree(tmpdir) 
```

This would clean up the code by a bit and follow best practice for pytest.

sappelhoff:
closes #55 

- replaces our custom tmpdir hacking with the pytest tmpdir fixture
- uses pytest.mark.parametrize for one test to make it (1) more concise, (2) more informative if failed
- "switches on" the comma test, **wich will fail until https://github.com/mne-tools/mne-python/pull/8492 is merged**

I plan that this is the last PR before release 0.4 #60

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ __pycache__
 coverage.xml
 /coverage
 htmlcov/
+.vscode/
--- a/pybv/tests/test_bv_writer.py
+++ b/pybv/tests/test_bv_writer.py
@@ -10,10 +10,7 @@
 #
 # License: BSD (3-clause)
 
-import os
 from datetime import datetime, timezone
-from shutil import rmtree
-from tempfile import mkdtemp
 
 import mne
 import numpy as np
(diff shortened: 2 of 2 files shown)


cbrnr:
How does this work? You just add one parameter called `tmpdir` and that's it?

sappelhoff:
> How does this work? You just add one parameter called tmpdir and that's it?

yes, and then pytest creates the directory, available from the `pathlib.Path` called `tmpdir` and also deletes that dir after the run succeeds or fails. Super convenient

cbrnr:
I don't think it cleans up after itself though, because it keeps the three most recently created directories. We might want to change that to delete a temp directory immediately (see https://github.com/pytest-dev/pytest/issues/543#issuecomment-306644153).

sappelhoff:
> We might want to change that to delete a temp directory immediately

I thought that was a good idea and tried to add a `setup.cfg` like:

```
[tool:pytest]
tmp_keep=1
```

but it doesn't work, and this is why: 

https://github.com/pytest-dev/pytest/issues/543#issuecomment-475204800

:)

cbrnr:
Should've read it till the end I guess 😉. LGTM!

Issue #50, opened 2020-11-05
Feature: Allow pybv to scale to several units (accomodate different channel types)


sappelhoff:
Currently pybv only accepts a single unit as a parameter:

https://github.com/bids-standard/pybv/blob/76e8bea84d6594ca8e87f8fe367896167e43c641/pybv/io.py#L62-L65

and that unit is applied to all channels:

https://github.com/bids-standard/pybv/blob/76e8bea84d6594ca8e87f8fe367896167e43c641/pybv/io.py#L276-L281

It would be nice to be able to pass a n_channels long vector of units to pybv, to accommodate different channel types. 

For example an auxilliary channel in degrees Celsius, which would be in the unit C.

Note that channel-specific resolutions are already possible, and the proposed feature here is simply a logical extension of that behavior:

https://github.com/bids-standard/pybv/blob/76e8bea84d6594ca8e87f8fe367896167e43c641/pybv/io.py#L57-L61

Issue #63, opened 2020-11-08
better file management


sappelhoff:
We should add a "clean-up" function that makes sure that whenever pybv.write_brainvision is called and errors in the middle of the process, no files or folders are left over.

That is: A bad run must not produce any folders or files

I think currently we do produce a folder on bad runs.

PR #66, opened 2020-12-18
[MRG] Allows one to pass in a list of units to scale each channel separately

adam2392:
came up in https://github.com/mne-tools/mne-bids/issues/664
Closes: #50 

Also, partially addresses #63 by moving the creation of the directory after the input argument checks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,5 @@
 # simple makefile to simplify repetitive build env management tasks under posix
 
-# caution: testing won't work on windows, see README
-
 PYTHON ?= python
 PYTESTS ?= pytest
 
--- a/Makefile
+++ b/Makefile
@@ -5,38 +5,13 @@
 PYTHON ?= python
 PYTESTS ?= pytest
 
-all: clean inplace test
-
-clean-pyc:
-	find . -name "*.pyc" | xargs rm -f
-
-clean-so:
-	find . -name "*.so" | xargs rm -f
-	find . -name "*.pyd" | xargs rm -f
-
-clean-build:
-	rm -rf _build
-
-clean-ctags:
-	rm -f tags
-
-clean-cache:
-	find . -name "__pycache__" | xargs rm -rf
-
-clean-e:
-	find . -name "*-e" | xargs rm -rf
-
-clean: clean-build clean-pyc clean-so clean-ctags clean-cache clean-e
+all: inplace test
 
 inplace:
 	$(PYTHON) setup.py develop
 
-test: inplace check-manifest
-	rm -f .coverage
-	$(PYTESTS) pybv
-
 test-doc:
-	$(PYTESTS) --doctest-modules --doctest-ignore-import-errors mne_bids
+	$(PYTESTS) --doctest-modules --doctest-ignore-import-errors pybv
 
 flake:
 	@if command -v flake8 > /dev/null; then \
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,60 @@
+# simple makefile to simplify repetitive build env management tasks under posix
+
+# caution: testing won't work on windows, see README
+
+PYTHON ?= python
+PYTESTS ?= pytest
+
+all: clean inplace test
+
+clean-pyc:
+	find . -name "*.pyc" | xargs rm -f
+
+clean-so:
+	find . -name "*.so" | xargs rm -f
+	find . -name "*.pyd" | xargs rm -f
+
+clean-build:
+	rm -rf _build
+
+clean-ctags:
+	rm -f tags
+
+clean-cache:
+	find . -name "__pycache__" | xargs rm -rf
+
+clean-e:
+	find . -name "*-e" | xargs rm -rf
+
+clean: clean-build clean-pyc clean-so clean-ctags clean-cache clean-e
+
+inplace:
+	$(PYTHON) setup.py develop
+
+test: inplace check-manifest
+	rm -f .coverage
+	$(PYTESTS) pybv
+
+test-doc:
+	$(PYTESTS) --doctest-modules --doctest-ignore-import-errors mne_bids
+
+flake:
+	@if command -v flake8 > /dev/null; then \
+		echo "Running flake8"; \
+		flake8 --docstring-convention numpy --count pybv; \
+	else \
+		echo "flake8 not found, please install it!"; \
+		exit 1; \
+	fi;
+	@echo "flake8 passed"
+
+pydocstyle:
+	@echo "Running pydocstyle"
+	@pydocstyle
+
+pep:
+	@$(MAKE) -k flake pydocstyle
+
+build-doc:
+	cd doc; make clean
+	cd doc; make html
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,9 +15,14 @@ Here we list a changelog of pybv.
 Current
 =======
 
+Changelog
+~~~~~~~~
+- :func:`pybv.write_brainvision` adds support for channels with non-Volt units, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_).
+- :func:`pybv.write_brainvision` automatically converts ``uV`` and ``μV`` (Greek μ) to ``µV`` (micro sign µ), by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_).
+
 API
 ~~~
-- :func:`pybv.write_brainvision` now accepts a list of units for the ``unit`` parameter, which will scale each channel separately. In addition, support for non-Volt channels are now supported for the ``data`` parameter by passing a list of str to ``unit``. In addition, ``uV`` now also converted to ``µV``, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
+- The ``unit`` parameter in :func:`pybv.write_brainvision` now accepts a list of units (one unit per channel), by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_).
 
 Authors
 ~~~~~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,7 +15,10 @@ Here we list a changelog of pybv.
 Current
 =======
 
-- Nothing yet
+API
+~~~
+- :func:`pybv.write_brainvision` now accepts a list of units in the ``unit`` keyword argument, which will scale each channel separately, by `Adam Li`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
+
 
 0.4.0
 =====
@@ -23,6 +26,10 @@ Current
 Changelog
 ~~~~~~~~~
 - Passing a "greek small letter mu" to the ``unit`` parameter in :func:`pybv.write_brainvision` instead of a "micro sign" is now permitted, because the former will be automatically convert to the latter, by `Stefan Appelhoff`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
+
+Authors
+~~~~~~~
+- `Adam Li`_
 
 Bug
 ~~~
@@ -119,6 +126,7 @@ Authors
 - `Phillip Alday`_
 - `Stefan Appelhoff`_
 - `Tristan Stenner`_
+- `Adam Li`_
 
 .. _Chris Holdgraf: https://bids.berkeley.edu/people/chris-holdgraf
 .. _Stefan Appelhoff: http://stefanappelhoff.com/
@@ -126,3 +134,4 @@ Authors
 .. _Phillip Alday: https://palday.bitbucket.io/
 .. _Clemens Brunner: https://cbrnr.github.io/
 .. _Richard Höchenberger: https://hoechenberger.net/
+.. _Adam Li: https://adam2392.github.io/
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,6 +19,9 @@ API
 ~~~
 - :func:`pybv.write_brainvision` now accepts a list of units in the ``unit`` keyword argument, which will scale each channel separately, by `Adam Li`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
 
+Authors
+~~~~~~~
+- `Adam Li`_
 
 0.4.0
 =====
@@ -26,10 +29,6 @@ API
 Changelog
 ~~~~~~~~~
 - Passing a "greek small letter mu" to the ``unit`` parameter in :func:`pybv.write_brainvision` instead of a "micro sign" is now permitted, because the former will be automatically convert to the latter, by `Stefan Appelhoff`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
-
-Authors
-~~~~~~~
-- `Adam Li`_
 
 Bug
 ~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,7 +17,7 @@ Current
 
 API
 ~~~
-- :func:`pybv.write_brainvision` now accepts a list of units for the ``unit`` parameter, which will scale each channel separately, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
+- :func:`pybv.write_brainvision` now accepts a list of units for the ``unit`` parameter, which will scale each channel separately. In addition, support for non-Volt channels are now supported for the ``data`` parameter by passing a list of str to ``unit``. In addition, ``uV`` now also converted to ``µV``, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
 
 Authors
 ~~~~~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,7 +17,7 @@ Current
 
 API
 ~~~
-- :func:`pybv.write_brainvision` now accepts a list of units in the ``unit`` keyword argument, which will scale each channel separately, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
+- :func:`pybv.write_brainvision` now accepts a list of units for the ``unit`` parameter, which will scale each channel separately, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
 
 Authors
 ~~~~~~~
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -81,10 +81,9 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
     unit : str | list of str
         The unit of the exported data. This can be one of 'V', 'mV', 'µV' (or
         equivalently 'uV') , or 'nV', which will scale the data accordingly.
-        Defaults to 'µV'. Can also be a list of units with
-        one unit per channel. Non-voltage channels are stored as is, for
-        example temperature might be available in ``°C``, which ``pybv`` will
-        not scale.
+        Defaults to 'µV'. Can also be a list of units with one unit per
+        channel. Non-voltage channels are stored as is, for example temperature
+        might be available in ``°C``, which ``pybv`` will not scale.
     fmt : str
         Binary format the data should be written as. Valid choices are
         'binary_float32' (default) and 'binary_int16'.
@@ -100,10 +99,10 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
     iEEG/EEG/MEG data is assumed to be in V, and we will scale these data
     to µV by default. Any unit besides µV is officially unsupported in the
     BrainVision specification. However, if one specifies other voltage units
-    such as 'mV' or 'nV', we will still scale the signals accordingly
-    in the exported file. We will also write channels with non-voltage units
-    such as ``°C`` as is (without scaling). For maximum compatibility,
-    all signals should be written as µV.
+    such as 'mV' or 'nV', we will still scale the signals accordingly in the
+    exported file. We will also write channels with non-voltage units such as
+    ``°C`` as is (without scaling). For maximum compatibility, all signals
+    should be written as µV.
 
     Example
     -------
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -40,10 +40,9 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
     Parameters
     ----------
     data : ndarray, shape (n_channels, n_times)
-        The raw data to export. Voltage data is assumed to be in
-        **Volts** and will be stored as specified by `unit`.
-        Other unsupported non-volt units are able to be specified
-        in ``unit`` parameter.
+        The raw data to export. Voltage data is assumed to be in **Volts** and
+        will be scaled as specified by ``unit``. Non-voltage channels (as
+        specified by ``unit``) are never scaled (e.g. '°C').
     sfreq : int | float
         The sampling frequency of the data.
     ch_names : list of strings, shape (n_channels,)
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -15,6 +15,7 @@ import datetime
 import os
 import warnings
 from os import path as op
+from warnings import warn
 
 import numpy as np
 
@@ -255,8 +256,10 @@ def _scale_data_to_unit(data, units):
         scale = SUPPORTED_UNITS.get(unit, None)
         if scale is None:
             msg = (f'Encountered unsupported unit: {unit}\n'
-                   f'Use one of the following: {set(SUPPORTED_UNITS.keys())}')
-            raise ValueError(msg)
+                   f'We are not scaling channel {idx}. If this is a mistake, '
+                   f'use one of the following: {set(SUPPORTED_UNITS.keys())}')
+            warn(msg)
+            scale = 1  # don't scale the data at all
         scales[idx] = scale
     return data * scales
 
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -40,8 +40,10 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
     Parameters
     ----------
     data : ndarray, shape (n_channels, n_times)
-        The raw data to export. Data is assumed to be in **Volts** and will be
-        stored as specified by `unit`.
+        The raw data to export. Voltage data is assumed to be in
+        **Volts** and will be stored as specified by `unit`.
+        Other unsupported non-volt units are able to be specified
+        in ``unit`` parameter.
     sfreq : int | float
         The sampling frequency of the data.
     ch_names : list of strings, shape (n_channels,)
@@ -99,6 +101,15 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
     If one passes in other voltage units, we will scale them accordingly,
     and we will also write unsupported units such as ``°C`` as is
     without scaling.
+
+    Example
+    -------
+    >>> data = np.random.rand((3, 5))
+    >>> # write data with varying units
+    >>> write_brainvision(data, sfreq=1, ch_names=['A1', 'A2', 'TEMP'],
+    >>>                   folder_out='./',
+    >>>                   fname_base='test_file',
+    >>>                   unit=['µV', 'µV', '°C'])
     """
     # Input checks
     ev_err = ("events must be an ndarray of shape (n_events, 2) or "
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -81,7 +81,7 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
         Defaults to 'µV'. Can also be a list of units with one unit per
         channel, where any arbitrary channel unit can be defined.
         For example, temperature might be ``°C``, which ``pybv``
-        will simply not scale..
+        will not scale.
     fmt : str
         Binary format the data should be written as. Valid choices are
         'binary_float32' (default) and 'binary_int16'.
@@ -94,9 +94,11 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
 
     Notes
     -----
-    Passing a list of arbitrary ``units`` should be double-checked. If
-    one mis-spells the supported voltage unit, then it will not get
-    scaled properly.
+    Passing a list of arbitrary ``units`` should be double-checked.
+    Any unit besides µV is unsupported in the BrainVision specification.
+    If one passes in other voltage units, we will scale them accordingly,
+    and we will also write unsupported units such as ``°C`` as is
+    without scaling.
     """
     # Input checks
     ev_err = ("events must be an ndarray of shape (n_events, 2) or "
@@ -259,18 +261,26 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date):
 
 def _scale_data_to_unit(data, units):
     """Scale `data` in Volts to `data` in `units`."""
+    # keep track of any unsupported units to warn user
+    unsupported_units = set()
+
     # create a vector to multiply with to play nice with numpy
     scales = np.zeros((len(units), 1))
     for idx, unit in enumerate(units):
         scale = SUPPORTED_UNITS.get(unit, None)
         if scale is None:
-            msg = (f'Encountered unsupported unit: {unit}\n'
-                   f'We are not scaling channel {idx}. If this is a mistake, '
-                   f'use one of the following: '
-                   f'{", ".join(sorted(SUPPORTED_UNITS.keys()))}')
-            warn(msg)
+            unsupported_units.add(unit)
             scale = 1  # don't scale the data at all
         scales[idx] = scale
+
+    if len(unsupported_units) > 0:
+        msg = (f'Encountered unsupported units: '
+               f'{", ".join(unsupported_units)}\n'
+               f'We are not scaling these channels. '
+               f'If this is a mistake, '
+               f'use one of the following: '
+               f'{", ".join(sorted(SUPPORTED_UNITS.keys()))}')
+        warn(msg)
     return data * scales
 
 
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -76,9 +76,8 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
         (``1e-6 * 0.1``).
     unit : str | list of strings
         The unit of the exported data. This can be one of 'V', 'mV', 'µV' (or
-        equivalently 'uV') , or 'nV'. Defaults to 'µV'. If a list is passed in,
-        it specifies the unit per channel. Then it must have the same length
-        as the number of channels.
+        equivalently 'uV') , or 'nV'. Defaults to 'µV'. Can also be a
+        list of units with one unit per channel.
     fmt : str
         Binary format the data should be written as. Valid choices are
         'binary_float32' (default) and 'binary_int16'.
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -74,7 +74,7 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
         resolution, and writing data in µV with 0.1 resolution will thus
         guarantee accurate writing vor all values >= 1e-7 µV
         (``1e-6 * 0.1``).
-    unit : str | list of strings, shape (n_channels,)
+    unit : str | list of strings
         The unit of the exported data. This can be one of 'V', 'mV', 'µV' (or
         equivalently 'uV') , or 'nV'. Defaults to 'µV'. If a list is passed in,
         it specifies the unit per channel. Then it must have the same length
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -268,7 +268,9 @@ def _scale_data_to_unit(data, units):
     scales = np.zeros((len(units), 1))
     for idx, unit in enumerate(units):
         scale = SUPPORTED_UNITS.get(unit, None)
-        if scale is None:
+        # unless the unit is 'µV', then
+        # technically it is not supported in BrainVision
+        if scale is None or unit != 'µV':
             unsupported_units.add(unit)
             scale = 1  # don't scale the data at all
         scales[idx] = scale
@@ -276,7 +278,6 @@ def _scale_data_to_unit(data, units):
     if len(unsupported_units) > 0:
         msg = (f'Encountered unsupported units: '
                f'{", ".join(unsupported_units)}\n'
-               f'We are not scaling these channels. '
                f'If this is a mistake, '
                f'use one of the following: '
                f'{", ".join(sorted(SUPPORTED_UNITS.keys()))}')
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -95,7 +95,7 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
     Notes
     -----
     Passing a list of arbitrary ``units`` should be double-checked.
-    Any unit besides µV is "unsupported" in the BrainVision specification.
+    Any unit besides µV is unsupported in the BrainVision specification.
     If one passes in other voltage units, we will scale them accordingly,
     and we will also write unsupported units such as ``°C`` as is
     without scaling.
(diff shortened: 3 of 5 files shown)


adam2392:
cc @pmeyers16

docs/changelog.rst
```diff
@@ -15,7 +15,10 @@ Here we list a changelog of pybv.
 Current
 =======
 
-- Nothing yet
+API
+~~~
+- :func:`pybv.write_brainvision` now accepts a list of units in the ``unit`` keyword argument, which will scale each channel separately, by `Adam Li`_ (`#47 <https://github.com/bids-standard/pybv/pull/47>`_)
```
adam2392, line 20:
```suggestion
- :func:`pybv.write_brainvision` now accepts a list of units in the ``unit`` keyword argument, which will scale each channel separately, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
```

adam2392:
@sappelhoff once this is merged, what are your thoughts on releasing 0.5? :p 

and then we can make the corresponding fix in mne-bids for different channels?

sappelhoff:
we can remove 0.5 directly after - but we need to properly review first :-) ... I'll find time soon, but perhaps @hoechenberger or @cbrnr can also give it a pass

pybv/io.py
```diff
@@ -74,9 +74,11 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
         resolution, and writing data in µV with 0.1 resolution will thus
         guarantee accurate writing vor all values >= 1e-7 µV
         (``1e-6 * 0.1``).
-    unit : str
+    unit : str | list of strings, shape (n_channels,)
```
hoechenberger, line 77:
```suggestion
    unit : str | list of strings
```
Lists don't have a shape, only a length

pybv/io.py
```diff
@@ -133,14 +129,27 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
-    if unit == 'μV':
-        # this is greek mu: μ
-        # https://www.compart.com/de/unicode/U+03BC
-        warnings.warn(
-            f"Encountered small greek letter mu: 'μ' in unit: {unit} ... "
-            f"converting to micro sign: 'µ': {unit.replace('μ', 'µ')}"
-        )
-        unit = 'µV'
+    # check unit is single str or list
+    if isinstance(unit, list):
```
hoechenberger, line 133: I wouldn't do this. Python uses duck typing, so I one should be able to expect that passing a tuple or numpy array should work as well.

Instead, I would _first_ check for `str` identity, and only if it's not a string, cast to a list via `list(unit)`, and then proceed.

hoechenberger (changes requested):

pybv/io.py
adam2392: Updated in https://github.com/bids-standard/pybv/pull/66/commits/0d920ea736c64cb50e2917eb0b7dcb5a8df40624

hoechenberger (approved):

docs/changelog.rst
```diff
@@ -15,7 +15,13 @@ Here we list a changelog of pybv.
 Current
 =======
 
-- Nothing yet
+API
+~~~
+- :func:`pybv.write_brainvision` now accepts a list of units in the ``unit`` keyword argument, which will scale each channel separately, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
```
sappelhoff, line 20:
```suggestion
- :func:`pybv.write_brainvision` now accepts a list of units for the ``unit`` parameter, which will scale each channel separately, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
```

pybv/io.py
```diff
@@ -74,9 +74,11 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
         resolution, and writing data in µV with 0.1 resolution will thus
         guarantee accurate writing vor all values >= 1e-7 µV
         (``1e-6 * 0.1``).
-    unit : str
+    unit : str | list of strings
```
sappelhoff, line 77: I prefer it if use actual types here, so str instead of string or strings ... and int instead of integer, etc.

```suggestion
    unit : str | list of str
```

pybv/io.py
```diff
@@ -74,9 +74,11 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
         The unit of the exported data. This can be one of 'V', 'mV', 'µV' (or
-        equivalently 'uV') , or 'nV'. Defaults to 'µV'.
+        equivalently 'uV') , or 'nV'. Defaults to 'µV'. If a list is passed in,
+        it specifies the unit per channel. Then it must have the same length
+        as the number of channels.
```
sappelhoff, line 81:
```suggestion
        equivalently 'uV') , or 'nV'. Defaults to 'µV'. Can also be a
        list of units with one unit per channel.
```

pybv/io.py
```diff
@@ -133,14 +129,27 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
+
+    # check units for compatibility with greek lettering
+    for idx, unit in enumerate(units):
+        if unit == 'μV':
+            # this is greek mu: μ
+            # https://www.compart.com/de/unicode/U+03BC
+            warnings.warn(
+                f"Encountered small greek letter mu: 'μ' in unit: {unit} ... "
+                f"converting to micro sign: 'µ': {unit.replace('μ', 'µ')}"
+            )
```
sappelhoff, line 150: IMHO, this warning should be sent once in that loop. I wouldn't want to see this warning nchans times as a user. Especially not if I originally just passed a single unit :-)

sappelhoff (dismissed):
LGTM so far. For all unicode symbols that you added to show "micro", I expect that you used the "true" micro symbol, and not the greek small letter mu? please just double check - there characters are hard to review via GitHub.

My main point is that this PR does not really close #50 completely. You *do* allow for passing a list of units, and that's cool (thanks a lot). But the scope of #50 was to also let users put in other units that may be used in AUX channels, such as degree celsius, etc.

Either way, I do not require you to implement that as well - I think your enhancement can be merged with the small improvements I suggested. Especially, if it solves the problem that you have  reported in https://github.com/mne-tools/mne-bids/issues/664 (does it?).

So either:

1. you make the adjustements, we merge, and you update issue #50 with some description of what remains to be done
2. you make some more modifications to this PR to truly close #50. Units that could/should be supported with pybv are those that are mentioned in the specification:

screenshots:

grafik
grafik


... and in addition the ones that MNE-Python can read:

https://github.com/mne-tools/mne-python/blob/8204c90211bf6a0d9385cd9e8f5fe41741a54ea2/mne/io/brainvision/brainvision.py#L529-L531

If I see it correctly, that'd be only `C` for degrees celsius on top of the Volt units. Is that correct @hoechenberger @cbrnr ?

But I am not sure whether we should support `C` actually ... `°C` or `oC` seems more sensible to me. 🤷  given this debate we could also skip `C` for now.

Thinking about it some more ... I am not sure about these uV² etc. either. What are your thoughts on this?

The idea with #50  was that a PR closing #50 should enable pybv to also write more heterogeneous data - For example when channel1 is in Volts, channel2 is in °C, and channel3 is in arbitrary units. (Currently all channels of the data to be written MUST be in volts)

PS: This ended up to be a train of thought ... sorry if the reading experience is unpleasant.

sappelhoff:
It seems like I also wrote #50 very badly so what I describe in my comment above what #50 was intended to mean, actually does not come across while re-reading.

adam2392:
@sappelhoff sorry  I'm a bit confused on certain things you said. Do you mind clarifying:

1. what are the units of misc channels? I assumed that this PR would handle it, but then I realized, perhaps these channels are not encoded as Volts...
2. Besides temperature (i.e. Celcius), what are other units? uV/m^2? Where are those documented in mne-python just so I can have a reference?

If it turns out that it's not too much of a rabbit hole... happy to implement :). You know me.

hoechenberger (approved):

hoechenberger:
> But I am not sure whether we should support `C` actually ... `°C` or `oC` seems more sensible to me. 🤷 given this debate we could also skip `C` for now.

Please don't use `C`, as this is the SI symbol for `Coulomb`. I think °C is fine, no?

cbrnr:
I think we should warn if we encounter a user-defined unit (i.e. a unit not defined in the [specs](file:///Users/clemens/Downloads/BVCDF-1.0_DocVersion-2.5_2019-12-12.pdf)).

Known units are: `μV`, `μV/Hz`, `μV²`, `μV²/Hz`, `μV/m²`.

We should make sure to use the official spelling, which includes the use of mu vs. micro and the right symbol for the superscript 2. Everything else is user-defined, so I don't think we should treat `C` or `°C` differently than other user-defined units.

sappelhoff:
> what are the units of misc channels? I assumed that this PR would handle it, but then I realized, perhaps these channels are not encoded as Volts...

I don't know ... I think they could be all kinds of units and MNE doesn't say? That could be a question for the MNE gitter channel

> Besides temperature (i.e. Celcius), what are other units? uV/m^2? Where are those documented in mne-python just so I can have a reference?

looking at the [BrainVision IO code in MNE](https://github.com/mne-tools/mne-python/blob/8204c90211bf6a0d9385cd9e8f5fe41741a54ea2/mne/io/brainvision/brainvision.py#L529-L531), I don't think any of those units mentioned in the BrainVision spec are supported in MNE.

> Please don't use C, as this is the SI symbol for Coulomb. I think °C is fine, no?

ugh! Why does MNE read C as °C :sob: 

https://github.com/mne-tools/mne-python/blob/8204c90211bf6a0d9385cd9e8f5fe41741a54ea2/mne/io/brainvision/brainvision.py#L303-L313

```python
_unit_dict = {'V': 1.,  # V stands for Volt
              'µV': 1e-6,
              'uV': 1e-6,
              'mV': 1e-3,
              'nV': 1e-9,
              'C': 1,  # C stands for celsius
              'µS': 1e-6,  # S stands for Siemens
              'uS': 1e-6,
              'ARU': 1,  # ARU is the unity for the breathing data
              'S': 1,
              'N': 1}  # Newton
```


> I think we should warn if we encounter a user-defined unit (i.e. a unit not defined in the specs).

mmmh generally agreed, but according to that, mV, nV, V, etc. are also user defined, and only µV is recognized ... which I think is unreasonable.

---

Overall, this is highly confusing. Let's get back to the issue at hand.

1. Users have problems converting their data to BrainVision format when the units that are contained differ from Volts.
2. This PR currently allows users to have different channels in different units, for example ECG in mV, and EEG channels in µV.

Potential solution: 

- provide an API for users to specify the units of INPUT data (i.e., for the [`data` parameter](https://pybv.readthedocs.io/en/stable/generated/pybv.write_brainvision.html#pybv.write_brainvision))
- provide a sensible (and easily extensible) selection of units that the data can be written for (i.e., for the `unit` parameter). For example, we can start with common things like `°C` etc.

We would then have to implement **reading** support for the pybv supported units in MNE and preferably also other libraries. 

Alternative solution:

Contact BrainProducts and ask them how they recommend users to specify units beyond the ones they mention in the spec.

sappelhoff:
actually, perhaps a third, pragmatic option would be to not "support" any units but to allow something like:

```python
# some data
n_chans = 3
n_times = 1000
data = np.random.randn((n_chans, n_times))

# the third channel is some misc channel
unit = ["µV", "mV", "°C"]

# pass to write_brainvision
# for chan 1 and 2, a "valid" unit is detected (µV and mV respectively)
# so they are scaled and assumed to be in Volts within `data`
# ... however, for chan 3, the unit is not recognized (°C),
# so chan 3 is assumed to NOT be in Volts --> no scaling is applied, data is written "as is"
# in VHDR file, °C is written as unit for chan 3
```

WDYT?

adam2392:
Kay now it's done:

New sample vhdr file looks like this: 
```
Brain Vision Data Exchange Header File Version 1.0
; Written using pybv 0.5.0.dev0

[Common Infos]
Codepage=UTF-8
DataFile=pybv.eeg
MarkerFile=pybv.vmrk
DataFormat=BINARY
; Data orientation: MULTIPLEXED=ch1,pt1, ch2,pt1 ...
DataOrientation=MULTIPLEXED
NumberOfChannels=10
; Sampling interval in microseconds
SamplingInterval=1000.0

[Binary Infos]
BinaryFormat=IEEE_FLOAT_32

[Channel Infos]
; Each entry: Ch<Channel number>=<Name>,<Reference channel name>,
; <Resolution in "Unit">,<Unit>, Future extensions..
; Fields are delimited by commas, some fields might be omitted (empty).
; Commas in channel names are coded as "\1".
Ch1=ch_0,,0.1,V
Ch2=ch_1,,0.1,V
Ch3=ch_2,,0.1,V
Ch4=ch_3,,0.1,V
Ch5=ch_4,,0.1,V
Ch6=ch_5,,0.1,V
Ch7=ch_6,,0.1,V
Ch8=ch_7,,0.1,V
Ch9=ch_8,,0.1,V
Ch10=ch_9,,0.1,°C

[Comment]
```

P.S. also added a Makefile to automate the pep/pydocstyle check.

adam2392:
Note that the one area that could run into problem is if people mis-spell mV, or like nV or something and things are not converted/scaled correctly.

Idk if people have any thoughts?

pybv/io.py
```diff
@@ -255,8 +256,10 @@ def _scale_data_to_unit(data, units):
         scale = SUPPORTED_UNITS.get(unit, None)
         if scale is None:
             msg = (f'Encountered unsupported unit: {unit}\n'
-                   f'Use one of the following: {set(SUPPORTED_UNITS.keys())}')
-            raise ValueError(msg)
+                   f'We are not scaling channel {idx}. If this is a mistake, '
+                   f'use one of the following: {set(SUPPORTED_UNITS.keys())}')
```
hoechenberger, line 269: can we do something like (you'd have to adjust for correct line length):

```python
f'use one of the following: {', '.join(sorted(SUPPORTED_UNITS.keys()))}')
```
you don't need `set()` as dict keys need to be unique anyway.

hoechenberger:
> if people mis-spell mV, or like nV or something

How would you mis-spell "mV"..?

adam2392:
> > if people mis-spell mV, or like nV or something
> 
> How would you mis-spell "mV"..?

SAy as like mv? or MV?

cbrnr:
I don't understand what we mean when we talk about "supported units". According to the specs, only `µV`, `µV/Hz`, `µV²`, `µV²/Hz`, and `µV/m²` are known to the recorder. This means that we should try to stick with these units for maximum compatibility. Is this what we are currently doing?

What about the units defined in `SUPPORTED_UNITS = {'V': 1e0, 'mV': 1e3, 'µV': 1e6, 'uV': 1e6, 'nV': 1e9}`? Are these units that we write to the file? All of these except `µV` are undefined.

sappelhoff:
thanks for your continued work Adam :-)

> Say as like mv? or MV?

we could add a check that calls str.lower()  on each unit, and if we suspect a capitalization typo took place, we can send a warning.

For example if a unit "mv" is specified, we warn --> "did you mean mV?"

> I don't understand what we mean when we talk about "supported units". 

currently, these are the pybv supported units: https://github.com/bids-standard/pybv/blob/31baca05fc6e1870c72f37e42716fa4812e4ad75/pybv/io.py#L31

>  This means that we should try to stick with these units [the ones mentioned in the BV spec] for maximum compatibility. Is this what we are currently doing?

square brackets added by me.

yes, we are currently going beyond what the BrainVision **recorder** software, and BrainVision **analyzer** software "know" and support. That is, we support scalings beyond "µV", such as nano Volt, milli Volt, Volt, and even the "uV" ascii form of "µV".

Two thoughts:

1. We should convert "uV" to "µV" when we encounter it, and send a warning, similar to what we do here: https://github.com/bids-standard/pybv/blob/31baca05fc6e1870c72f37e42716fa4812e4ad75/pybv/io.py#L136-L142 ... I see no point to writing "uV", given that the VHDR file is utf-8, and Brain Products prescribes µ as the character


2. I think it's probably okay to also support nV, mV, and V on top of µV ... **BUT** we should check whether (1) fieldtrip, (2) EEGLAB, and (3) BrainStorm support this as well.

Makefile
```diff
@@ -0,0 +1,60 @@
...
+PYTESTS ?= pytest
+
+all: clean inplace test
+
+clean-pyc:
+	find . -name "*.pyc" | xargs rm -f
+
+clean-so:
+	find . -name "*.so" | xargs rm -f
+	find . -name "*.pyd" | xargs rm -f
```
sappelhoff, line 15: is this needed? what does it do?

Let's keep this Makefile minimal ... see also "clean-ctags", "clean-e", etc.

Please remove all commands that you do not need.

hoechenberger:
> we could add a check that calls str.lower() on each unit, and if we suspect a capitalization typo took place, we can send a warning.
> 
> For example if a unit "mv" is specified, we warn --> "did you mean mV?"
> 
> >

But… why? milli-Volts is mV, not mv or MV. Users should be able to get the capitalization right.

We could issue a warning if all units appear to be "user-defined", but beyond that, I wouldn't try to do anything clever.

hoechenberger:
> 2\. I think it's probably okay to also support nV, mV, and V on top of µV ... **BUT** we should check whether (1) fieldtrip, (2) EEGLAB, and (3) BrainStorm support this as well.

Why would we want to support something that is not part of the official file format spec? This is bound to cause issues, if you ask me…

sappelhoff:
>  I wouldn't try to do anything clever.

fine with me

> Why would we want to support something that is not part of the official file format spec?

I think @tstenner introduced that in order to optimize saving int16 data .. but I really don't remember, see also: https://github.com/bids-standard/pybv/issues/38

perhaps we should start only writing µV? Like ... deprecate the feature of writing nV, mV, V, ... uV?

cbrnr:
I don't remember either, but there were signals that benefitted from storing them in nV.

I suggest that we stick to the specs, which means µV (and yes, let's just replace uV and the other µV with the official µV).

Whenever we do something behind the scenes (e.g. replace uV with µV), we should issue a warning. Similarly, if the user chooses a unit such as nV, we should issue a warning that this unit is not officially supported.

hoechenberger:
@cbrnr Sounds good to me!

cbrnr:
@adam2392 this looks good now! However, can you also add a warning if a (supported) unit other than µV is used? The note is useful, but I think we should also warn.

Makefile
```diff
@@ -0,0 +1,35 @@
+# simple makefile to simplify repetitive build env management tasks under posix
+
+# caution: testing won't work on windows, see README
```
sappelhoff, line 3:
```suggestion
```

pybv/io.py
```diff
@@ -86,13 +91,15 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
-    # Create output file names/paths
-    os.makedirs(folder_out, exist_ok=True)
-    vhdr_fname = op.join(folder_out, fname_base + '.vhdr')
-    vmrk_fname = op.join(folder_out, fname_base + '.vmrk')
-    eeg_fname = op.join(folder_out, fname_base + '.eeg')
 
+    Notes
+    -----
+    Passing a list of arbitrary ``units`` should be double-checked.
+    Any unit besides µV is "unsupported" in the BrainVision specification.
```
cbrnr, line 98: No need to put "unsupported" in quotes.

pybv/io.py
```diff
@@ -133,14 +140,31 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
+                         f"does not match number of channel "
+                         f"names ({nchan})")
+    units = unit
+
+    # check units for compatibility with greek lettering
+    show_warning = False
+    for idx, unit in enumerate(units):
+        if unit == 'μV':
+            # this is greek mu: μ
+            # https://www.compart.com/de/unicode/U+03BC
```
cbrnr, line 158: Move the two lines of comment to line 156 and shorten to `Greek mu μ (U+03BC)`.

pybv/io.py
```diff
@@ -133,14 +140,31 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
+            unit = 'µV'
```
cbrnr, line 159: Add a comment `micro symbol µ (U+00B5)`.

pybv/io.py
```diff
@@ -133,14 +140,31 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
+    if len(unit) != nchan:
+        raise ValueError(f"Number of channels in unit ({len(unit)}) "
+                         f"does not match number of channel "
+                         f"names ({nchan})")
+    units = unit
+
+    # check units for compatibility with greek lettering
+    show_warning = False
+    for idx, unit in enumerate(units):
+        if unit == 'μV':
```
cbrnr, line 156: Here we also want to replace `uV`.

pybv/io.py
```diff
@@ -133,14 +140,31 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
+            # this is greek mu: μ
+            # https://www.compart.com/de/unicode/U+03BC
+            unit = 'µV'
+            units[idx] = unit
+            show_warning = True
+    # only show the warning once if a greek letter was encountered
+    if show_warning:
```
cbrnr, line 163: The message needs to be adapted because we now also replace `uV`.

pybv/io.py
```diff
@@ -228,19 +259,36 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date):
...
+            scale = 1  # don't scale the data at all
+        scales[idx] = scale
+
+    if len(unsupported_units) > 0:
+        msg = (f'Encountered unsupported units: '
+               f'{", ".join(unsupported_units)}\n'
+               f'If this is a mistake, '
+               f'use one of the following: '
+               f'{", ".join(sorted(SUPPORTED_UNITS.keys()))}')
+        warn(msg)
```
cbrnr, line 286: Could we differentiate between truly unsupported units (which we don't scale) and internally supported (but officially unsupported) units (i.e. `V`, `mV`, `nV`) and issue a different warning message? We do scale those volts in contrast to truly unsupported units.

cbrnr (changes requested):

pybv/io.py
adam2392: Done with a different message.

docs/changelog.rst
```diff
@@ -15,7 +15,13 @@ Here we list a changelog of pybv.
 Current
 =======
 
-- Nothing yet
+API
+~~~
+- :func:`pybv.write_brainvision` now accepts a list of units for the ``unit`` parameter, which will scale each channel separately, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
```
sappelhoff, line 20: I think we can also add a new point regarding the fact that we also support non Volt channels now for the `data` parameter, and that these are "detected" through passing a list of str to `unit`, where the respective entries differ from nV, mV, µV, uV, or V.

Also another point about uV now being auto converted to µV.

pybv/io.py
```diff
@@ -74,9 +75,13 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
         resolution, and writing data in µV with 0.1 resolution will thus
         guarantee accurate writing vor all values >= 1e-7 µV
         (``1e-6 * 0.1``).
-    unit : str
+    unit : str | list of str
```
sappelhoff, line 78: the text under the `data` parameter needs to be modified as well, because also non-Volt data is now "allowed" as well, but the interplay is a bit complex with the `unit` param, so we need to brainstorm for a good phrasing.

Potentially it'd be good to also add a short example to the README or the docstring (if feasible in only a few lines)

docs/changelog.rst
```diff
@@ -15,7 +15,13 @@ Here we list a changelog of pybv.
 Current
 =======
 
-- Nothing yet
+API
+~~~
+- :func:`pybv.write_brainvision` now accepts a list of units for the ``unit`` parameter, which will scale each channel separately. In addition, support for non-Volt channels are now supported for the ``data`` parameter by passing a list of str to ``unit``. In addition, ``uV`` now also converted to ``µV``, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_)
```
cbrnr, line 20:
```suggestion
Changelog
~~~~~~~~
- :func:`pybv.write_brainvision` adds support for channels with non-Volt units, by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_).
- :func:`pybv.write_brainvision` automatically converts ``uV`` and ``μV`` (Greek μ) to ``µV`` (micro sign µ), by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_).

API
~~~
- The ``unit`` parameter in :func:`pybv.write_brainvision` now accepts a list of units (one unit per channel), by `Adam Li`_ (`#66 <https://github.com/bids-standard/pybv/pull/66>`_).
```

hoechenberger (approved):

pybv/io.py
```diff
@@ -39,8 +40,10 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
     Parameters
     ----------
     data : ndarray, shape (n_channels, n_times)
-        The raw data to export. Data is assumed to be in **Volts** and will be
-        stored as specified by `unit`.
+        The raw data to export. Voltage data is assumed to be in
+        **Volts** and will be stored as specified by `unit`.
+        Other unsupported non-volt units are able to be specified
+        in ``unit`` parameter.
```
cbrnr, line 46:
```suggestion
        The raw data to export. Voltage data is assumed to be in **Volts** and
        will be scaled as specified by ``unit``. Non-voltage channels (as
        specified by ``unit``) are never scaled (e.g. '°C').
```

pybv/io.py
```diff
@@ -74,9 +77,13 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
         (``1e-6 * 0.1``).
-    unit : str
+    unit : str | list of str
         The unit of the exported data. This can be one of 'V', 'mV', 'µV' (or
-        equivalently 'uV') , or 'nV'. Defaults to 'µV'.
+        equivalently 'uV') , or 'nV', which will then scale the data to 'V'.
+        Defaults to 'µV'. Can also be a list of units with one unit per
+        channel, where any arbitrary channel unit can be defined.
+        For example, temperature might be ``°C``, which ``pybv``
+        will not scale.
```
cbrnr, line 86:
```suggestion
        equivalently 'uV') , or 'nV', which will scale the data (assumed to be
        in 'V') accordingly. Defaults to 'µV'. Can also be a list of units with
        one unit per channel. Non-voltage channels are stored as is, for
        example temperature might be available in ``°C``, which ``pybv`` will
        not scale.
```

pybv/io.py
hoechenberger: Shouldn't this be 

    `'°C'`

(with backticks)

pybv/io.py
```diff
@@ -86,13 +93,24 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
-    vmrk_fname = op.join(folder_out, fname_base + '.vmrk')
-    eeg_fname = op.join(folder_out, fname_base + '.eeg')
 
+    Notes
+    -----
+    Passing a list of arbitrary ``units`` should be double-checked.
+    Any unit besides µV is unsupported in the BrainVision specification.
+    If one passes in other voltage units, we will scale them accordingly,
+    and we will also write unsupported units such as ``°C`` as is
+    without scaling.
```
cbrnr, line 103:
```suggestion
    EEG/MEG data is assumed to be in V, and we will scale these data to µV by
    default. Any unit besides µV is unsupported in the BrainVision
    specification. However, if one specifies other voltage units such as 'mV'
    or 'nV', we will still scale the signals accordingly in the exported file. ,
    We will also write channels with unsupported units such as ``°C`` as is
    (without scaling). For maximum compatibility, all signals should be written
    as µV.
```

pybv/io.py
```diff
@@ -86,13 +93,24 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
+
+    Example
+    -------
+    >>> data = np.random.rand((3, 5))
+    >>> # write data with varying units
+    >>> write_brainvision(data, sfreq=1, ch_names=['A1', 'A2', 'TEMP'],
+    >>>                   folder_out='./',
+    >>>                   fname_base='test_file',
+    >>>                   unit=['µV', 'µV', '°C'])
```
cbrnr, line 112:
```suggestion
    >>>                   unit=['µV', 'mV', '°C'])
```

pybv/io.py
```diff
@@ -133,14 +151,33 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
+            show_warning = True
+
+    # only show the warning once if a greek letter was encountered
+    if show_warning:
         warnings.warn(
-            f"Encountered small greek letter mu: 'μ' in unit: {unit} ... "
-            f"converting to micro sign: 'µ': {unit.replace('μ', 'µ')}"
+            f"Encountered small greek letter mu: 'μ' in unit, or the "
+            f"letter u: {unit} ... "
+            f"converting to micro sign: 'µ'"
```
cbrnr, line 179:
```suggestion
            f"Encountered small Greek letter mu 'μ' or 'u' in unit: {unit}. "
            f"Converting to micro sign 'µ'."
```

pybv/io.py
```diff
@@ -228,19 +272,51 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date):
...
-    return data * scale
+def _scale_data_to_unit(data, units):
+    """Scale `data` in Volts to `data` in `units`."""
+    # keep track of any unsupported units to warn user
+    # these are voltage units that we convert to microvolts
+    unsupported_units = set()
+
+    # officially unsupported units that contain non-voltage
+    # related signals
+    official_unsupp_units = set()
```
cbrnr, line 283: Can you rename these to `voltage_units` and `non_voltage_units`?

cbrnr: Officially unsupported and unsupported is not really easy to understand (so maybe also remove this terminology from the comments).

pybv/io.py
```diff
@@ -228,19 +272,51 @@ def _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date):
...
+        elif scale is None:
+            official_unsupp_units.add(unit)
+            scale = 1  # don't scale the data at all
+        scales[idx] = scale
+
+    if len(unsupported_units) > 0:
+        msg = (f'Encountered unsupported voltage units: '
+               f'{", ".join(unsupported_units)}\n'
+               f'We will automatically scale and convert '
+               f'these to Volts. If this is a mistake, '
```
cbrnr, line 303: This is not correct, we will not scale to Volts, but to the unit given (e.g. 'mV'). The original data must be in Volts.

pybv/io.py
cbrnr: I don't know, every project uses this differently. Backticks will render italic, whereas double backticks will render as monospace. Not sure what we want.

pybv/io.py
```diff
@@ -40,10 +40,9 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
     Parameters
     ----------
     data : ndarray, shape (n_channels, n_times)
-        The raw data to export. Voltage data is assumed to be in
-        **Volts** and will be stored as specified by `unit`.
-        Other unsupported non-volt units are able to be specified
-        in ``unit`` parameter.
+        The raw data to export. Voltage data is assumed to be in **Volts** and
+        will be scaled as specified by ``unit``. Non-voltage channels (as
+        specified by ``unit``) are never scaled (e.g. '°C').
```
hoechenberger, line 45:
```suggestion
        specified by ``unit``) are never scaled (e.g. ``'°C'``).
```

hoechenberger:
@adam2392 btw you can also apply an entire stack of suggestions at once by going to the "Files Changed" pane and using the "Add suggestion to batch" button for all suggestions you'd like to apply.

pybv/io.py
adam2392: Done.

adam2392: Lmk wyt?

adam2392:
> @adam2392 btw you can also apply an entire stack of suggestions at once by going to the "Files Changed" pane and using the "Add suggestion to batch" button for all suggestions you'd like to apply.

Ah nice. that's a good feature. Thanks ;)

pybv/io.py
```diff
@@ -28,7 +29,9 @@
 
 SUPPORTED_ORIENTS = {'multiplexed'}
 
-SUPPORTED_UNITS = {'V': 1e0, 'mV': 1e3, 'µV': 1e6, 'uV': 1e6, 'nV': 1e9}
+SUPPORTED_VOLTAGE_UNITS = {
```
hoechenberger, line 32: I'm wondering if this should be renamed to `SUPPORTED_VOLTAGE_SCALINGS`?

adam2392: I like that better.

pybv/tests/test_bv_writer.py
```diff
@@ -97,18 +97,22 @@ def test_bv_bad_format(tmpdir):
...
     with pytest.raises(ValueError, match='Orientation bad not supported'):
         _write_bveeg_file(eeg_fname, data=data, orientation='bad',
-                          format="bad", resolution=1e-6, unit="µV")
+                          format="bad", resolution=1e-6,
+                          units=["µV"] * n_chans)
     with pytest.raises(ValueError, match='Data format bad not supported'):
         _write_bveeg_file(eeg_fname, data=data, orientation='multiplexed',
-                          format="bad", resolution=1e-6, unit="µV")
+                          format="bad", resolution=1e-6,
+                          units=["µV"] * n_chans)
```
cbrnr, line 115: Why these additional arguments? Don't the original test work anymore?

pybv/tests/test_bv_writer.py
```diff
@@ -246,3 +251,76 @@ def test_sampling_frequencies(tmpdir, sfreq):
...
+                      unit=units)
+    raw_written = mne.io.read_raw_brainvision(vhdr_fname=vhdr_fname,
+                                              preload=True)
+
+    # Check that the correct units were written in the BV file
+    orig_units = [u for key, u in raw_written._orig_units.items()]
+    assert len(set(orig_units)) == 2
+    assert all([orig_units[idx] == unit.replace("u", "µ") for idx in range(5)])
+    assert all([orig_units[-idx] == other_unit.replace("u", "µ")
+                for idx in range(1, 6)])
```
cbrnr, line 299: Maybe replace `range(5)` and `range(1, 6)` with the actual channel number (i.e. `n_chans`)?

pybv/tests/test_bv_writer.py
```diff
@@ -246,3 +251,76 @@ def test_sampling_frequencies(tmpdir, sfreq):
...
+                                              preload=True)
+
+    # check round-trip works
+    absolute_tolerance = 0
+    assert_allclose(data, raw_written.get_data(), atol=absolute_tolerance)
+
+    # Check that the correct units were written in the BV file
+    orig_units = [u for key, u in raw_written._orig_units.items()]
+    assert len(set(orig_units)) == 2
+    assert all([orig_units[idx] == unit for idx in range(9)])
```
cbrnr, line 325: Same here, use `n_chans` instead of magic number 9.

pybv/tests/test_bv_writer.py
```diff
@@ -246,3 +251,76 @@ def test_sampling_frequencies(tmpdir, sfreq):
     raw_written = mne.io.read_raw_brainvision(vhdr_fname=vhdr_fname,
                                               preload=True)
     assert_allclose(sfreq, raw_written.info['sfreq'])
+
+
+@pytest.mark.parametrize("unit", SUPPORTED_VOLTAGE_UNITS)
+def test_write_multiple_units(tmpdir, unit):
```
cbrnr, line 257: Do we test writing a file with multiple channels using a `str` argument for `unit` (e.g. `unit='mV'`)? Most tests now pass a list, but we should also make sure that passing a string continues to work. If this is covered, then ignore this comment (I only looked at your changes and not the full file).

pybv/io.py
```diff
@@ -74,9 +78,13 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
         (``1e-6 * 0.1``).
-    unit : str
+    unit : str | list of str
         The unit of the exported data. This can be one of 'V', 'mV', 'µV' (or
-        equivalently 'uV') , or 'nV'. Defaults to 'µV'.
+        equivalently 'uV') , or 'nV', which will scale the data accordingly.
+        Defaults to 'µV'. Can also be a list of units with
+        one unit per channel. Non-voltage channels are stored as is, for
+        example temperature might be available in ``°C``, which ``pybv`` will
+        not scale.
```
cbrnr, line 87: Can you re-wrap according to PEP8?

pybv/io.py
```diff
@@ -86,13 +94,26 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
 
+    Notes
+    -----
+    iEEG/EEG/MEG data is assumed to be in V, and we will scale these data
+    to µV by default. Any unit besides µV is officially unsupported in the
+    BrainVision specification. However, if one specifies other voltage units
+    such as 'mV' or 'nV', we will still scale the signals accordingly
+    in the exported file. We will also write channels with non-voltage units
+    such as ``°C`` as is (without scaling). For maximum compatibility,
+    all signals should be written as µV.
```
cbrnr, line 106: This looks like it could also be re-wrapped to 79 characters.

pybv/tests/test_bv_writer.py
adam2392: Passing in a single string to the private function is not supported anymore (and the tests will fail) because it expects a list of units, which is set in the public API.

pybv/tests/test_bv_writer.py
adam2392: Yep! this is still tested in:

```
@pytest.mark.parametrize("format", SUPPORTED_FORMATS.keys())
@pytest.mark.parametrize("resolution", resolutions)
@pytest.mark.parametrize("unit", SUPPORTED_VOLTAGE_SCALINGS)
def test_format_resolution_unit(tmpdir, format, resolution, unit):
    """Test different combinations of formats, resolutions, and units."""
    # Check whether this test will be numerically possible
    tmpdata = _scale_data_to_unit(data.copy(), [unit] * n_chans)
    tmpdata = tmpdata * np.atleast_2d((1 / resolution)).T
    _, dtype = _chk_fmt(format)
    data_will_fit = _check_data_in_range(tmpdata, dtype)

    kwargs = dict(data=data, sfreq=sfreq, ch_names=ch_names,
                  fname_base=fname, folder_out=tmpdir,
                  resolution=resolution, unit=unit,
                  fmt=format)
```

adam2392: Yep! this is tested in the following function:

```
@pytest.mark.parametrize("format", SUPPORTED_FORMATS.keys())
@pytest.mark.parametrize("resolution", resolutions)
@pytest.mark.parametrize("unit", SUPPORTED_VOLTAGE_SCALINGS)
def test_format_resolution_unit(tmpdir, format, resolution, unit):
    """Test different combinations of formats, resolutions, and units."""
    # Check whether this test will be numerically possible
    tmpdata = _scale_data_to_unit(data.copy(), [unit] * n_chans)
    tmpdata = tmpdata * np.atleast_2d((1 / resolution)).T
    _, dtype = _chk_fmt(format)
    data_will_fit = _check_data_in_range(tmpdata, dtype)

    kwargs = dict(data=data, sfreq=sfreq, ch_names=ch_names,
                  fname_base=fname, folder_out=tmpdir,
                  resolution=resolution, unit=unit,
                  fmt=format)

    if not data_will_fit:
        # End this test early
        match = f"can not be represented in '{format}' given"
        with pytest.raises(ValueError, match=match):
            write_brainvision(**kwargs)
        return

    write_brainvision(**kwargs)
```

pybv/tests/test_bv_writer.py
adam2392: The public API still supports a single string

cbrnr (approved):

cbrnr:
Thanks for your patience @adam2392, I think this is now ready (provided all tests pass). @hoechenberger @sappelhoff could you give it a final look?

hoechenberger (approved):

sappelhoff (approved):
Great work, thanks @adam2392 and @hoechenberger + @cbrnr for reviews!

adam2392:
Yay! @sappelhoff 0.5 release possibly?

I can make a few corresponding PRs into the mne-bids now to update to allow arbitrary units that are supported by mne-python.

sappelhoff:
> Yay! @sappelhoff 0.5 release possibly?

yes, I'll look into it with #67 also, and possibly changing from "authors each release" to "authors overall", like in [pyprep](https://pyprep.readthedocs.io/en/latest/whats_new.html#id39)

> I can make a few corresponding PRs into the mne-bids now to update to allow arbitrary units that are supported by mne-python.

that'd be great - it'll allow us to have a final "practical check" of the new feature before release.

PR #70, opened 2021-01-04
Improve Makefile

### LINKED ISSUE #69: Simplify Makefile rule?
I'm not a Makefile expert, but could the rule

```
pep:
	@$(MAKE) -k flake pydocstyle check-manifest
```

be simplified to

```
pep: flake pydocstyle check-manifest
```

? If not I'd be curious to learn why.

Furthermore, I think it is common practice to define phony targets explicitly, i.e. the first rule should be

```
.PHONY: all inplace test flake pydocstyle check-manifest pep build-doc
```

cbrnr:
Fixes #69. I tested `make pep` on my local machine and it is still working.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,12 @@
-# simple makefile to simplify repetitive build env management tasks under posix
-
-PYTHON ?= python
-PYTESTS ?= pytest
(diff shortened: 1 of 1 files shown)


sappelhoff (approved):
Thank you @cbrnr

PR #76, opened 2021-08-02
Add CITATION.cff file

sappelhoff:
see also: https://github.com/mne-tools/mne-python/issues/9624

What are your opinions?

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,2 @@
+Chris Holdgraf <choldgraf@gmail.com>
+Chris Holdgraf <choldgraf@gmail.com> <choldgraf@berkeley.edu>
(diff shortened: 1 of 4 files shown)


cbrnr:
Looks good! What is the purpose of the `.mailmap` file?

sappelhoff:
> What is the purpose of the .mailmap file?

It's so that `git shortlog --summary --email` gives a reasonable output. And we can use that output to track authors (concretely: people who have committed code). Else it would show multiple entries for one and the same author if that author happened to have diverging git configs at different points in time.

sappelhoff:
Initially I had removed phillip from the author list, but after sleeping over it I reconsidered that it's better to instead explicitly add him in a commit (using GitHub's "Co-authored-by" feature) and to the author list. The initial port of the code was based on his work, so I think this it's better this way.

I edited this squashed-and-merged PR's commit (`git commit --amend`) and force pushed to `main`:

https://github.com/bids-standard/pybv/commit/e68158de07775bcfec7e6dcd64f51bce86f54145

PR #78, opened 2021-08-06
[MRG] Improve file handling, add overwrite parameter

### LINKED ISSUE #63: better file management

sappelhoff:
closes #63 

- better file handling: Delete intermediate files when writing fails
- better input checks: raise error when ch_names is not list of str (or list of int which we can easily turn into str)
- add `overwrite` parameter and set new default (False) --> previously pybv *always* overwrote files, which was a bit dangerous IMHO

Also adds tests for all of those

### DIFF
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -14,8 +14,10 @@
 import codecs
 import datetime
 import os
+import shutil
 import warnings
 from os import path as op
+from pathlib import Path
 from warnings import warn
 
 import numpy as np
@@ -203,21 +205,26 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
                          '"YYYYMMDDhhmmssuuuuuu".')
 
     # Create output file names/paths
+    folder_out_created = not op.exists(folder_out)
     os.makedirs(folder_out, exist_ok=True)
-
     vhdr_fname = op.join(folder_out, fname_base + '.vhdr')
     vmrk_fname = op.join(folder_out, fname_base + '.vmrk')
     eeg_fname = op.join(folder_out, fname_base + '.eeg')
 
-    # Write output files
-    # NOTE: call _write_bveeg_file first, so that if it raises ValueError,
-    # no files are written.
-    _write_bveeg_file(eeg_fname, data, orientation='multiplexed', format=fmt,
-                      resolution=resolution, units=units)
-    _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date)
-    _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq,
-                     ch_names, orientation='multiplexed', format=fmt,
-                     resolution=resolution, units=units)
+    # Write output files, but delete everything if we come across an error
+    try:
+        _write_bveeg_file(eeg_fname, data, orientation='multiplexed',
+                          format=fmt, resolution=resolution, units=units)
+        _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date)
+        _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq,
+                         ch_names, orientation='multiplexed', format=fmt,
+                         resolution=resolution, units=units)
+    except ValueError:
+        for fname in (eeg_fname, vmrk_fname, vhdr_fname):
+            Path(fname).unlink(missing_ok=True)
+        if folder_out_created:
+            shutil.rmtree(folder_out)
+        raise
 
 
 def _chk_fmt(fmt):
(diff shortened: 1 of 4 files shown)


pybv/io.py
```diff
@@ -36,6 +36,7 @@
 
 
 def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
+                      overwrite=False,
                       events=None, resolution=0.1, unit='µV',
                       fmt='binary_float32', meas_date=None):
```
cbrnr, line 41: What about using Black-ish formatting here, which will make it easier to see the parameters?

```
def write_brainvision(
    *, data, sfreq, ch_names, fname_base, folder_out,
    overwrite=False,
    events=None,
    resolution=0.1,
    unit='µV',
    fmt='binary_float32',
    meas_date=None
):

File: pybv/io.py [hunk 2]
```diff
@@ -125,6 +128,9 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
 
     """
     # Input checks
+    if not isinstance(overwrite, bool) and overwrite not in [0, 1]:
```
cbrnr, line 131: I would only allow `True` and `False` here. What's the point of allowing also 0 and 1? IMO it is irrelevant whether or not Python represents `True` and `False` as `int` internally.

pybv/io.py
```diff
@@ -140,6 +146,10 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
             raise ValueError(ev_err)
 
     nchan = len(ch_names)
+    for ch in ch_names:
+        if not isinstance(ch, (str, int)):
+            raise ValueError("ch_names must be a list of str.")
```
cbrnr, line 151: If it can be a list of `int` I would mention it.

pybv/io.py
```diff
@@ -202,22 +210,37 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
...
-    _write_vmrk_file(vmrk_fname, eeg_fname, events, meas_date)
-    _write_vhdr_file(vhdr_fname, vmrk_fname, eeg_fname, data, sfreq,
-                     ch_names, orientation='multiplexed', format=fmt,
-                     resolution=resolution, units=units)
+    vmrk_fname = op.join(folder_out, fname_base + '.vmrk')
+    vhdr_fname = op.join(folder_out, fname_base + '.vhdr')
+    msg = "File already exists: {}.\nConsider setting overwrite=True."
+    for fname in (eeg_fname, vmrk_fname, vhdr_fname):
+        if op.exists(fname) and not overwrite:
+            raise IOError(msg.format(fname))
```
cbrnr, line 222:
```suggestion
    for fname in (eeg_fname, vmrk_fname, vhdr_fname):
        if op.exists(fname) and not overwrite:
            raise IOError(f"File already exists: {fname}.\n"
                          f"Consider setting overwrite=True."))
```

cbrnr (commented):
In addition to my specific comments I was wondering if `os.path` could also be replace by `pathlib` not just in the tests but also in the rest of the code?

pybv/io.py
sappelhoff: I had this because many people use 0 and 1 as a shorthand for False and True, and isinstance wouldn't accept that, see:

```
In [1]: isinstance(1, bool)
Out[1]: False
```

sappelhoff:
> In addition to my specific comments I was wondering if os.path could also be replace by pathlib not just in the tests but also in the rest of the code?

Yes, I thought about that myself. Would probably be prettier/clearer. Feel free to open an issue that then somebody can do at some point.

pybv/io.py
cbrnr: Yeah, I don't like it. They should be using `True` and `False`.

sappelhoff: okay --> eefd93ef5a37808439883af561f6dd9bcb46ac5b

pybv/io.py
```diff
@@ -49,7 +49,7 @@ def write_brainvision(*, data, sfreq, ch_names, fname_base, folder_out,
         specified by ``unit``) are never scaled (e.g. ``'°C'``).
     sfreq : int | float
         The sampling frequency of the data.
-    ch_names : list of strings, shape (n_channels,)
+    ch_names : list of str, shape (n_channels,)
```
hoechenberger: unrelated by why do we keep writing "shape ..." for a list; it should only have a length…

sappelhoff: :+1: 

https://github.com/bids-standard/pybv/pull/78/commits/efbb1fc4de93b0262d943ed19cc0f3714cb2905a

cbrnr:
Thanks @sappelhoff!

PR #80, opened 2021-09-24
Force little-endian format when writing eeg datafile

### LINKED ISSUE #79: Tests fail with big-endian machines
Hello Devs

Thanks a lot for providing pybv, it is a very useful tool that is used in neuroimaging. I am trying to include python-pybv for Fedora, a Linux distribution. I was trying to run the tests, but they fail on a big-endian machine, s390x architecture to be precise. Below is a snippet of a failing test:

```bash
___________________ test_comma_in_ch_name[ch_names_tricky1] ____________________
tmpdir = local('/tmp/pytest-of-mockbuild/pytest-0/test_comma_in_ch_name_ch_names1')
ch_names_tricky = ['ch_0,foo', 'ch_1,foo', 'ch_2,foo', 'ch_3,foo', 'ch_4,foo', 'ch_5,foo', ...]
    @requires_version("mne", min_version="0.22")
    @pytest.mark.parametrize("ch_names_tricky",
                             [[ch + ' f o o' for ch in ch_names],
                              [ch + ',foo' for ch in ch_names],
                              ]
                             )
    def test_comma_in_ch_name(tmpdir, ch_names_tricky):
        """Test that writing channel names with special characters works."""
        # write and read data to BV format
        write_brainvision(data=data, sfreq=sfreq, ch_names=ch_names_tricky,
                          fname_base=fname, folder_out=tmpdir)
        vhdr_fname = tmpdir / fname + '.vhdr'
        raw_written = mne.io.read_raw_brainvision(vhdr_fname=vhdr_fname,
                                                  preload=True)
    
        assert ch_names_tricky == raw_written.ch_names  # channels
    
>       assert_allclose(data, raw_written._data)  # data round-trip
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       x and y nan location mismatch:
E        x: array([[-7.031873e-06, -4.902824e-06, -3.218143e-06, ..., -4.052640e-06,
E               -5.178996e-06,  1.410489e-06],
E              [ 1.635612e-05,  1.622443e-05,  4.016866e-06, ..., -5.808766e-06,...
E        y: array([[ 4.759927e-16, -4.722928e+18, -6.062177e-01, ...,  4.668447e-41,
E               -8.139047e+19, -1.879789e-24],
E              [-1.627288e-18, -5.858951e-38, -1.282439e-13, ..., -3.478547e-04,...
pybv/tests/test_bv_writer.py:147: AssertionError
```

The tests work fine on the little-endian machines, which basically implies that the bug might be where/how we are storing the files. I tried going through pybv's io and from what I could understand it is writing the file in the UTF-8 format, therefore the endianness [shouldn't matter](https://stackoverflow.com/a/23831655).

However, when I went through mne's brainvision io, I could see that they [open the file as a binary file](https://github.com/mne-tools/mne-python/blob/main/mne/io/brainvision/brainvision.py#L75) which could be causing the tests to fail.

I am not sure if this is the case, but if it is I would also file a ticket at mne so that we could get a resolution and support more architectures. If you could look into this and let us know it would be great!

Aniket-Pradhan:
Also updated the docs to add information about this change 😸 

closes #79

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -34,7 +34,7 @@ Changelog
 ~~~~~~~~~
 
 - :func:`pybv.write_brainvision` gained a new parameter, ``ref_ch_names``, to specify the reference channels used during recording, by `Richard Höchenberger`_ and `Stefan Appelhoff`_ (:gh:`75`)
-- Fix bug where :func:`pybv.write_brainvision` would write the binary eeg file in big-endian on a big-endian system. This could lead to non-uniformity of data-files written by big-endian or little-endian machines, by `Aniket Pradhan`, `Clemens Brunner` and `Stefan Appelhoff`_ (:gh:`80`).
+- Fix bug where :func:`pybv.write_brainvision` would write the binary eeg file in big-endian on a big-endian system. This could lead to non-uniformity of data-files written by big-endian or little-endian machines, by `Aniket Pradhan`_, `Clemens Brunner`_ and `Stefan Appelhoff`_ (:gh:`80`).
 
 API
 ~~~
@@ -127,3 +127,4 @@ Changelog
 .. _Clemens Brunner: https://cbrnr.github.io/
 .. _Richard Höchenberger: https://hoechenberger.net/
 .. _Adam Li: https://adam2392.github.io/
+.. _Aniket Pradhan: http://home.iiitd.edu.in/~aniket17133/
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,7 @@ Authors
 People who contributed to this software across releases (in **alphabetical order**):
 
 - `Adam Li`_
+- `Aniket Pradhan`_
 - `Chris Holdgraf`_
 - `Clemens Brunner`_
 - `Phillip Alday`_
@@ -33,6 +34,7 @@ Changelog
 ~~~~~~~~~
 
 - :func:`pybv.write_brainvision` gained a new parameter, ``ref_ch_names``, to specify the reference channels used during recording, by `Richard Höchenberger`_ and `Stefan Appelhoff`_ (:gh:`75`)
+- Fix bug where :func:`pybv.write_brainvision` would write the binary eeg file in big-endian on a big-endian system. This could lead to non-uniformity of data-files written by big-endian or little-endian machines, by `Aniket Pradhan`, `Clemens Brunner` and `Stefan Appelhoff`_ (:gh:`80`).
 
 API
 ~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -32,13 +32,15 @@ Current (unreleased)
 
 Changelog
 ~~~~~~~~~
-
 - :func:`pybv.write_brainvision` gained a new parameter, ``ref_ch_names``, to specify the reference channels used during recording, by `Richard Höchenberger`_ and `Stefan Appelhoff`_ (:gh:`75`)
-- Fix bug where :func:`pybv.write_brainvision` would write the binary file in big-endian on a big-endian system, by `Aniket Pradhan`_, `Clemens Brunner`_, and `Stefan Appelhoff`_ (:gh:`80`)
 
 API
 ~~~
 - :func:`pybv.write_brainvision` now has an ``overwrite`` parameter that defaults to ``False``, by `Stefan Appelhoff`_ (:gh:`78`).
+
+Bug
+~~~
+- Fix bug where :func:`pybv.write_brainvision` would write the binary file in big-endian on a big-endian system, by `Aniket Pradhan`_, `Clemens Brunner`_, and `Stefan Appelhoff`_ (:gh:`80`)
 
 0.5.0 (2021-01-03)
 ==================
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -34,7 +34,7 @@ Changelog
 ~~~~~~~~~
 
 - :func:`pybv.write_brainvision` gained a new parameter, ``ref_ch_names``, to specify the reference channels used during recording, by `Richard Höchenberger`_ and `Stefan Appelhoff`_ (:gh:`75`)
-- Fix bug where :func:`pybv.write_brainvision` would write the binary eeg file in big-endian on a big-endian system. This could lead to non-uniformity of data-files written by big-endian or little-endian machines, by `Aniket Pradhan`_, `Clemens Brunner`_ and `Stefan Appelhoff`_ (:gh:`80`).
+- Fix bug where :func:`pybv.write_brainvision` would write the binary file in big-endian on a big-endian system, by `Aniket Pradhan`_, `Clemens Brunner`_, and `Stefan Appelhoff`_ (:gh:`80`)
 
 API
 ~~~
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -494,8 +494,8 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
 
     # Swap bytes if system architecure is big-endian and dtype is "native" or
     # "big-endian"
-    if ((sys.byteorder == "big" and data.dtype.byteorder == "=") or
-            data.dtype.byteorder == ">"):
+    byteorder = data.dtype.byteorder
+    if (sys.byteorder == "big" and byteorder == "=") or byteorder == ">":
         data = data.byteswap()
 
     # Save to binary
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -492,11 +492,13 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
         raise ValueError(msg)
     data = data.astype(dtype=dtype)
 
-    # Swap bytes if system architecture is big-endian and dtype is "native" or
-    # "big-endian"
-    byteorder = data.dtype.byteorder
-    if ((sys.byteorder == "big" and byteorder == "=") or
-            byteorder == ">"):  # pragma: no cover
+    # We always write data as little-endian without BOM
+    # `data` is already in native byte order due to numpy operations that
+    # result in copies of the `data` array (see above)
+    assert data.dtype.byteorder == "="
+
+    # swap bytes if system architecture is big-endian
+    if sys.byteorder == "big":  # pragma: no cover
         data = data.byteswap()
 
     # Save to binary
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -494,8 +494,8 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
 
     # Swap bytes if system architecure is big-endian and dtype is "native" or
     # "big-endian"
-    if (sys.byteorder == "big" and
-       (data.dtype.byteorder == "=" or data.dtype.byteorder == ">")):
+    if ((sys.byteorder == "big" and data.dtype.byteorder == "=") or
+            data.dtype.byteorder == ">"):
         data = data.byteswap()
 
     # Save to binary
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -494,8 +494,8 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
 
     # Swap bytes if system architecure is big-endian and dtype is "native" or
     # "big-endian"
-    byteorder = data.dtype.byteorder  # pragma: no cover
-    if (sys.byteorder == "big" and byteorder == "=") or byteorder == ">":  # pragma: no cover  # noqa: E501
+    byteorder = data.dtype.byteorder
+    if (sys.byteorder == "big" and byteorder == "=") or byteorder == ">":  # pragma: no cover noqa: E501
         data = data.byteswap()
 
     # Save to binary
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -495,7 +495,8 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
     # Swap bytes if system architecure is big-endian and dtype is "native" or
     # "big-endian"
     byteorder = data.dtype.byteorder
-    if (sys.byteorder == "big" and byteorder == "=") or byteorder == ">":  # pragma: no cover noqa: E501
+    if ((sys.byteorder == "big" and byteorder == "=") or
+            byteorder == ">"):  # pragma: no cover
         data = data.byteswap()
 
     # Save to binary
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -492,7 +492,7 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
         raise ValueError(msg)
     data = data.astype(dtype=dtype)
 
-    # Swap bytes if system architecure is big-endian and dtype is "native" or
+    # Swap bytes if system architecture is big-endian and dtype is "native" or
     # "big-endian"
     byteorder = data.dtype.byteorder
     if ((sys.byteorder == "big" and byteorder == "=") or
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -14,6 +14,7 @@
 import datetime
 import os
 import shutil
+import sys
 from os import path as op
 from warnings import warn
 
@@ -490,6 +491,12 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
             msg += "\nPlease consider writing using 'binary_float32' format."
         raise ValueError(msg)
     data = data.astype(dtype=dtype)
+
+    # Swap bytes if system architecure is big-endian and dtype is "native" or
+    # "big-endian"
+    if (sys.byteorder == "big" and
+       (data.dtype.byteorder == "=" or data.dtype.byteorder == ">")):
+        data = data.byteswap()
 
     # Save to binary
     data.ravel(order='F').tofile(eeg_fname)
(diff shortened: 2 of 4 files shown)


cbrnr (approved):

Aniket-Pradhan:
Force pushed to fix the flake8 tests. Should work fine now

pybv/io.py
```diff
@@ -491,5 +492,11 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
         raise ValueError(msg)
     data = data.astype(dtype=dtype)
 
+    # Swap bytes if system architecure is big-endian and dtype is "native" or
+    # "big-endian"
+    if (sys.byteorder == "big" and
+       (data.dtype.byteorder == "=" or data.dtype.byteorder == ">")):
+        data = data.byteswap()
```
sappelhoff, line 499: I don't really get this if clause.

Shouldn't we swap bytes if `sys.byteorder == "big"` AND `data.dtype.byteorder == "="` ... OR `data.dtype.byteorder == ">"`

In your case, couldn't it be possible that we are on a little-endian machine the byte order is big-endian? --> then big-endian data would be written, because the if-clause does not get triggered?

@Aniket-Pradhan

cbrnr: Theoretically this could happen, so this should probably be changed to:

```
if (sys.byteorder == "big" and data.dtype.byteorder == "=") or data.dtype.byteorder == ">":
        data = data.byteswap()
```

cbrnr: OTOH, when will `data.dtype.byteorder == ">"` ever be `True` on a little Endian machine? The default is always `=` AFAIK, so checking for `sys.byteorder` might be completely sufficient, i.e.

```
if sys.byteorder == "big":
    data = data.byteswap()
```

Aniket-Pradhan: It's not usually ideal that the byteorder would be big-endian on a little-endian machine, but yes, we can swap the bytes for this scenario. I'll update this in a new commit.

Aniket-Pradhan: > OTOH, when will `data.dtype.byteorder == ">"` ever be `True` on a little Endian machine? The default is always `=` AFAIK, so checking for `sys.byteorder` might be completely sufficient, i.e.

This is true, but one cannot be too sure of what the user is passing `data` as. That's why I added in the checks. It may be redundant, but we can be confident that we have covered all the possible cases.

pybv/io.py
```diff
@@ -491,5 +492,11 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
         raise ValueError(msg)
     data = data.astype(dtype=dtype)
 
+    # Swap bytes if system architecure is big-endian and dtype is "native" or
+    # "big-endian"
+    if ((sys.byteorder == "big" and data.dtype.byteorder == "=") or
+            data.dtype.byteorder == ">"):
+        data = data.byteswap()
```
cbrnr, line 499: If you want to avoid indentation you could define `byteorder = data.dtype.byteorder`, then the `if` condition fits in one line.

```suggestion
    byteorder = data.dtype.byteorder
    if (sys.byteorder == "big" and byteorder == "=") or byteorder == ">":
        data = data.byteswap()
```

sappelhoff (commented):
Cool, thanks both. Then only three things remain:

1. @Aniket-Pradhan please add an entry to the changelog here: https://github.com/bids-standard/pybv/blob/main/docs/changelog.rst, copying the style of the other entries (and add yourself to the list of authors)
2. Please append your name and info to https://github.com/bids-standard/pybv/blob/main/CITATION.cff as well
3. please add a `# pragma: no cover` (or whatever it was that prevents code coverage analysis to account for the lines) to the if clause, because we are (deliberately) not covering that in our tests, and I don't want coverage to get lower due to deliberate decisions :-)

docs/changelog.rst
```diff
@@ -33,6 +34,7 @@ Changelog
 ~~~~~~~~~
 
 - :func:`pybv.write_brainvision` gained a new parameter, ``ref_ch_names``, to specify the reference channels used during recording, by `Richard Höchenberger`_ and `Stefan Appelhoff`_ (:gh:`75`)
+- Fix bug where :func:`pybv.write_brainvision` would write the binary eeg file in big-endian on a big-endian system. This could lead to non-uniformity of data-files written by big-endian or little-endian machines, by `Aniket Pradhan`, `Clemens Brunner` and `Stefan Appelhoff`_ (:gh:`80`).
```
sappelhoff, line 37:
```suggestion
- Fix bug where :func:`pybv.write_brainvision` would write the binary eeg file in big-endian on a big-endian system. This could lead to non-uniformity of data-files written by big-endian or little-endian machines, by `Aniket Pradhan`_, `Clemens Brunner`_ and `Stefan Appelhoff`_ (:gh:`80`).
```

sappelhoff: please also see line 129 in this file, where you need to add a line like this one:

`.. _Adam Li: https://adam2392.github.io/`

pybv/io.py
```diff
@@ -494,8 +494,8 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
 
     # Swap bytes if system architecure is big-endian and dtype is "native" or
     # "big-endian"
-    byteorder = data.dtype.byteorder
-    if (sys.byteorder == "big" and byteorder == "=") or byteorder == ">":
+    byteorder = data.dtype.byteorder  # pragma: no cover
+    if (sys.byteorder == "big" and byteorder == "=") or byteorder == ">":  # pragma: no cover  # noqa: E501
```
sappelhoff, line 498: The first pragma is not needed (that line is covered), and I removed one `#` which I think is also not needed

```suggestion
    byteorder = data.dtype.byteorder
    if (sys.byteorder == "big" and byteorder == "=") or byteorder == ">":  # pragma: no cover noqa: E501
```

cbrnr: Why don't we cover this with a test? We just need to create a `data` array with `dtype=">f8"` or something.

Aniket-Pradhan: ohhh, sure. I'll write a small test for it as well

Aniket-Pradhan: Also, let me know what the merging strategy is. I could squash and tidy up the commits if you want.

docs/changelog.rst
```diff
@@ -34,7 +34,7 @@ Changelog
 ~~~~~~~~~
 
 - :func:`pybv.write_brainvision` gained a new parameter, ``ref_ch_names``, to specify the reference channels used during recording, by `Richard Höchenberger`_ and `Stefan Appelhoff`_ (:gh:`75`)
-- Fix bug where :func:`pybv.write_brainvision` would write the binary eeg file in big-endian on a big-endian system. This could lead to non-uniformity of data-files written by big-endian or little-endian machines, by `Aniket Pradhan`, `Clemens Brunner` and `Stefan Appelhoff`_ (:gh:`80`).
+- Fix bug where :func:`pybv.write_brainvision` would write the binary eeg file in big-endian on a big-endian system. This could lead to non-uniformity of data-files written by big-endian or little-endian machines, by `Aniket Pradhan`_, `Clemens Brunner`_ and `Stefan Appelhoff`_ (:gh:`80`).
```
cbrnr, line 37:
```suggestion
- Fix bug where :func:`pybv.write_brainvision` would write the binary file in big-endian on a big-endian system, by `Aniket Pradhan`_, `Clemens Brunner`_, and `Stefan Appelhoff`_ (:gh:`80`)
```

pybv/io.py
cbrnr: No worries, we'll squash before merging.

Aniket-Pradhan: Added in a small test. Also rebased the main branch of my fork to see if the travis tests work or not. They hopefully should work.

sappelhoff (approved):
awesome, thanks! I'll wait for a note from you that everything did indeed pass ... then we can merge and I'll make a new release.

Aniket-Pradhan:
There seems to be something wrong with the test, and I am guessing it is because we cannot pass a big-endian formatted array in a little-endian machine and expect it to carry out the scaling (and other arithmetic) operations correctly.

For example, if we reverse the bytes and change the dtype of a numpy array the scaling operation would give a different result on a little-endian and a big-endian machine.

Furthermore, `numpy.tofile` loses the endianness information of the numpy array and exports the data to whatever the machine's endianness is. The test was working earlier because I had accidentally passed the wrong variable, but upon reviewing the coverage report I found this mistake.

Therefore, I don't think it would be possible to test this section on a little-endian machine. OTOH, I guess we can replace `if (sys.byteorder == "big" and byteorder == "=") or byteorder == ">":` with `if sys.byteorder == "big":` because it would not make sense to work with big-endian dtype on a little-endian machine.

sappelhoff:
I see. In that case, let's still keep the check for endianness there - even if only 1 in 10k people run into it, it's a small thing to do and makes the code also easier to understand IMHO. 

Then we remove the test, and add the `pragma: no cover` back, to get a 100% diff coverage

cbrnr:
> There seems to be something wrong with the test, and I am guessing it is because we cannot pass a big-endian formatted array in a little-endian machine and expect it to carry out the scaling (and other arithmetic) operations correctly.

Wait, that's exactly what should happen. If you create a big-endian array, NumPy should know how to handle that data and perform operations correctly.

E.g. these two arrays are identical after multiplication:

```python
import numpy as np

little = np.arange(1e6, dtype="<f8").reshape(1000, 1000)
big = np.arange(1e6, dtype=">f8").reshape(1000, 1000)

little *= 7
big *= 7

np.array_equal(little, big)
```

If this example does not demonstrate the issue I'd be curious to see something that doesn't work.

cbrnr:
When writing arrays to disk with `.tofile()`, the files are not identical, which means that they should differ in their endianness:

```
little.tofile("little.dat")
big.tofile("big.dat")
!md5sum big.dat little.dat 
5157b19a956273b22d8cd056b5eea320  big.dat
d2ae4a99442623c5ef2678a0ea5a3429  little.dat
```

Aniket-Pradhan:
You're right, numpy is handling the operations correctly. I was also swapping the bytes along with changing the dtype, which now seems the wrong way to do it.

However, the byte order of the numpy array is reset to the native machine's endianness "`=`" whenever it is getting assigned (ref lines [476](https://github.com/bids-standard/pybv/blob/2b97fbc6e4c00e03df2d97c302566dd8f5da445d/pybv/io.py#L476), [480](https://github.com/bids-standard/pybv/blob/2b97fbc6e4c00e03df2d97c302566dd8f5da445d/pybv/io.py#L480) and [493](https://github.com/bids-standard/pybv/blob/2b97fbc6e4c00e03df2d97c302566dd8f5da445d/pybv/io.py#L493), where we set it manually from the [supported formats](https://github.com/bids-standard/pybv/blob/2b97fbc6e4c00e03df2d97c302566dd8f5da445d/pybv/io.py#L26)).

```python
import numpy as np

def scale_arr(arr, scale=10):
    return arr * scale

big = np.arange(1e6, dtype=">f8").reshape(1000, 1000)
big_scaled = scale_arr(big)

print(big.dtype, big.dtype.byteorder)
print(big_scaled.dtype, big_scaled.dtype.byteorder)
```

From what I can understand, I guess numpy only retains the datatype in the same scope of the function call.

Since the byte order is reset, the array is no longer little-endian, and hence the control doesn't go in the `if` condition, not allowing us to test the block, thereby reducing the coverage. This is why I had to revert to the old code because the coverage still dropped when I committed the test case the first time.

sappelhoff:
@cbrnr merge when you are happy.

cbrnr:
Indeed it seems like byte order is not preserved when copies are involved, e.g. `big = big * 7` will return a new array with native endianness but in-place operations such as `big *= 7` will preserve endianness. Therefore, I would try to change operations involving copies with in-place variants if possible. However, this will probably require a ton of modifications that I'm not sure we want to maintain. So I guess I'll probably merge, but I would like to confirm with @sappelhoff.

pybv/io.py
```diff
@@ -491,5 +492,12 @@ def _write_bveeg_file(eeg_fname, data, orientation, format, resolution, units):
         raise ValueError(msg)
     data = data.astype(dtype=dtype)
 
+    # Swap bytes if system architecure is big-endian and dtype is "native" or
```
sappelhoff, line 495:
```suggestion
    # Swap bytes if system architecture is big-endian and dtype is "native" or
```

sappelhoff:
> However, this will probably require a ton of modifications that I'm not sure we want to maintain. So I guess I'll probably merge, but I would like to confirm with @sappelhoff.

Yes, thanks for confirming with me - I also think we should merge. We clarified a major point in this issue/PR with Brain Products and made sure that pybv correctly writes as little-endian from now on. The only issue that remains would be a person on a little-endian machine creating a big-endian array, and then passing it to `write_brainvision` ... and I think it's safe to say that'll never happen (hehe).

cbrnr:
> ... and I think it's safe to say that'll never happen (hehe).

Famous last words. The only thing that is bugging me is that we don't have a test.

sappelhoff:
> Famous last words. The only thing that is bugging me is that we don't have a test.

yes, fair enough (that's also why I typed `hehe`). Could we add a check to the top of `write_brainvision` that errors when an array is not in native byte order? That way we would cover all cases, right? The only untested thing (`pragma: no cover`) would then be the if-clause to convert a native byte order array on a big-endian machine to little-endian --> and that's something we could test by adding a new CI on a big-endian architecture. (we discussed that previously but decided against it, but we can reconsider)

sappelhoff:
Okay, but that brings us back to your comment from https://github.com/bids-standard/pybv/pull/80#issuecomment-928925147

> Therefore, I would try to change operations involving copies with in-place variants if possible. However, this will probably require a ton of modifications that I'm not sure we want to maintain.

So we would need to make sure an array keeps its byte order as it passes through `write_brainvision`, see https://github.com/bids-standard/pybv/pull/80#issuecomment-928214651

If that's possible, then yes, we can test everything and ensure that passing any byte ordered array on any architecture results in (correctly scaled/processed) little endian format.

Aniket-Pradhan:
> So we would need to make sure an array keeps its byte order as it passes through `write_brainvision`, see [#80 (comment)](https://github.com/bids-standard/pybv/pull/80#issuecomment-928214651)

One shabby solution I was thinking, is that we could reset the dtype of `data` every time it is assigned a new copy. That way, we retain the appropriate dtype until we reach the final `if` condition. Although, that would be a lot of overhead for a very far-off case. Furthermore, we would have to use the appropriate dtype from the [supported formats](https://github.com/bids-standard/pybv/blob/2b97fbc6e4c00e03df2d97c302566dd8f5da445d/pybv/io.py#L26).

I am sure we can arrive at a much better solution than this, but I cannot think of anything better right now.

sappelhoff:
> One shabby solution I was thinking, is that we could reset the dtype of data every time it is assigned a new copy. 

It's pragmatic, I like that. And maybe some of the times we can just avoid making a copy and thus retain the dtype without the "shabby solution" :-)

> Furthermore, we would have to use the appropriate dtype from the supported formats.

you mentioned that before, could you please elaborate why that would be a problem?

cbrnr:
OK, so I looked at the changes again and it seems like we can simplify the condition to:

```python
if sys.byteorder == "big" and byteorder == "=":  # pragma: no cover
        data = data.byteswap()
```

The case `byteorder == ">"` never occurs because we're doing some copying before, which changes the array to native order.

Therefore, unless we change a lot of code just for the sake of testing on an LE architecture, we really have to mark this line as untestable. However, I think that's fine – sorry if this was clear for you previously, I thought we could somehow create a test that works on LE without changing too much.

We can always add this later if we think it is necessary.

sappelhoff (approved):

cbrnr:
Perfect! Thanks @Aniket-Pradhan and @sappelhoff!

sappelhoff:
@Aniket-Pradhan please run your tests and make sure everything passes :-) ... then notify me and I'll release pybv 0.6.0

Aniket-Pradhan:
ohh, looks like I am late to the party 😛 



> It's pragmatic, I like that. And maybe some of the times we can just avoid making a copy and thus retain the dtype without the "shabby solution" :-)

As @cbrnr pointed out, `data` would automatically be converted to the native architecture before it reaches the `if` condition. Therefore unless we want to modify a chunk of code just to test a possibility (which should be working anyways, because it is automatically converted to the native arch beforehand) doesn't seem feasible to me. 

That's why I was calling it a "shabby" solution, because it is not really needed, lol. Anyways, thanks for being supportive. Let me know if you plan to carry out with a similar solution, and I would be happy to help 😄 

> > Furthermore, we would have to use the appropriate dtype from the supported formats.
> 
> you mentioned that before, could you please elaborate why that would be a problem?

I basically meant that we would have to specify the endianness in the supported formats if we were to proceed with testing the code. 😛 

> @Aniket-Pradhan please run your tests and make sure everything passes :-) ... then notify me and I'll release pybv 0.6.0

I'll run the tests on travis, and hopefully it should work fine. I'll ping here as soon as the build finishes.

Thanks a lot @cbrnr and @sappelhoff for your help. You all rock!

Aniket-Pradhan:
@sappelhoff everything looks [green](https://app.travis-ci.com/github/Aniket-Pradhan/pybv). We are good to go! :D

[Build Status](https://app.travis-ci.com/Aniket-Pradhan/pybv)

sappelhoff:
https://pypi.org/project/pybv/0.6.0/ :tada: feel free to draft a tweet @Aniket-Pradhan or  @cbrnr  or @hoechenberger - I'll like and retweet of course :wink:  (but no energy to draft one myself)

PR #85, opened 2022-05-23
Add alternative software

### LINKED ISSUE #84: Add a list of alternative packages to documentation
There are tons of BrainVision readers and writers out there (I am trying to find some with regard to #83) ... it would probably be a good idea to list those in a documentation page (for Python, Matlab, Julia, ... ?).

- BrainVision Analyzer
- MNE-Python
- FieldTrip
- EEGLAB --> https://github.com/arnodelorme/bva-io
- BioSig
- ...

sappelhoff:
closes #84 

see: https://pybv--85.org.readthedocs.build/en/85/#alternatives


cbrnr (approved):
LGTM. It would be nice if the table had separate columns for e.g. read support, write support, open source – but it looks like there is not enough horizontal space in the docs layout, so the current version is completely fine!

PR #100, opened 2022-10-24
check ich "ch_names" present in annotations

### LINKED ISSUE #99: no 'ch_names' in annotations -> Error
I'm using pybv (v0.7.4) via MNE (v1.1.1) to concatenate and rewrite raw BV files. 
It raises a `Key error: ch_names` and stops. It seems the culprit is [here](https://github.com/bids-standard/pybv/blob/4af6bbab86e6151a3a64d759d4eca4a15e9b253a/pybv/_export.py#L134) in  [_export.py](https://github.com/bids-standard/pybv/blob/4af6bbab86e6151a3a64d759d4eca4a15e9b253a/pybv/_export.py). 
I checked my raw file in MNE, and its annotations do indeed not contain an entry for `ch_names`. Did not dig deeper into the MNE code to find out what `ch_names` mean in the context of annotations (not directly obvious to me). 
Is that an issue with my files or could [_export.py](https://github.com/bids-standard/pybv/blob/4af6bbab86e6151a3a64d759d4eca4a15e9b253a/pybv/_export.py) check for the existence of the key first and skip it if not present (dunno if this will break things downstream)?

eioe:
closes #99 .

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -32,6 +32,9 @@ authors:
 - given-names: "Pierre"
   family-names: "Cutellic"
   orcid: "https://orcid.org/0000-0002-7224-9222"
+- given-names: "Felix"
+  family-names: "Klotzsche"
+  orcid: "https://orcid.org/0000-0003-3985-2481"
 type: software
 repository-code: "https://github.com/bids-standard/pybv"
 url: "https://pybv.readthedocs.io/"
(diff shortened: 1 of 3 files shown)


sappelhoff (commented):
looks good to me, thanks!

Please:

1. add yourself to this file: https://github.com/bids-standard/pybv/blob/1a29f7413e6b8cffeb0caa6365443d0bd217735f/CITATION.cff#L32-L34
2. add an entry to the changelog, following the syntax you can see in previous entries: https://github.com/bids-standard/pybv/blob/main/docs/changelog.rst

eioe:
thx. Pls check my comment in the issue, before you merge. So we make sure, I did not introduce some new problems downstream.

sappelhoff:
I forgot to mention that you also need to add yourself here:

https://github.com/bids-standard/pybv/blob/1a29f7413e6b8cffeb0caa6365443d0bd217735f/docs/changelog.rst?plain=1#L9-L27

sappelhoff (approved):
Thanks for your first contribution @eioe -- may there be many more!

PR #115, opened 2024-02-11
Replace flake8/isort/black with ruff

cbrnr:
This PR replaces flake8/isort/black with ruff. In addition, I've (mostly manually) re-formatted docstrings so that they now use the line length limit of 88 chars. I've also removed `make pep`, because this is now being taken care of via pre-commit.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -39,7 +39,7 @@ you can run tests by making use of the `Makefile` and
 From the project root, call:
 
 - `make test` to run tests and coverage
-- `make pep` to run style checks
+- `pre-commit run` to run style checks (Ruff and some additional hooks)
 
 ## Building the documentation
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [tool.ruff.lint]
 select = ["A", "B006", "D", "E", "F", "I", "W", "UP"]
-ignore = ["A002", "D203", "D213"]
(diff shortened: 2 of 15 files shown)


pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[tool.ruff]
+exclude = ["docs"]
```
sappelhoff, line 2: why exclude docs? I'd be up for including this.

sappelhoff (approved):
Wonderful, this has been on my to do list for a long time. Thanks for taking this over @cbrnr. 

Just a small comment why not do docs/ as well (but could be done in a follow up, so I am approving anyhow).

pyproject.toml
cbrnr: Sure, done.

cbrnr (commented):
This is good to go IMO, please feel free to merge when green.

.github/CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ you can run tests by making use of the `Makefile` and
 From the project root, call:
 
 - `make test` to run tests and coverage
-- `make pep` to run style checks
+- `pre-commit run` to run style checks (Ruff and some additional hooks)
```
hoechenberger, line 42: Shouldn't this better be `pre-commit run -a`?

cbrnr: Yes, otherwise it will only run against staged files.

PR #116, opened 2024-02-11
Use pyproject.toml

cbrnr:
And while we're at it, we can also replace the legacy `setup.{cfg,py}` with `pyproject.toml`.


sappelhoff (approved):
LGTM

sappelhoff:
Thanks @cbrnr

PR #118, opened 2024-02-19
Support arbitrary markers

### LINKED ISSUE #103: Support strings for marker descriptions
Currently, we only support integer marker descriptions for "Stimulus" and "Response" marker types. String descriptions are only possible with "Comment" types.

This is a serious restriction, because it can lead to loss of information. I suggest that we also allow string descriptions for "Stimulus" and "Response" markers, which would allow us to just use the original (e.g. `mne.Annotations`) description.

Note that the specs do not contain our restriction at all. I already discussed this with my contact at BrainProducts, he says that this is a convention originated from their recorder software (which uses individual trigger lines). The format itself does not mandate integer marker descriptions at all.

cbrnr:
Fixes #103. This is a big (breaking) change, I decided to tackle this as follows:

- If the passed events is an `np.ndarray`, the previous conversion magic is still happening (i.e. `type` is set to `"Stimulus"` and integers are converted to `"S  1"` etc.).
- If the passed events is a list of dicts, markers are written as is.

This means that if someone needs to write "old-style" markers with custom types (such as `"Response"`, `"Comment"`, etc.), they must manually create a corresponding list of dicts.

I'm not sure if we need to change anything in MNE-Python, but I don't think so. It reads descriptions and types and combines them to "Type/Description", which I think should continue to work. The events from annotations heuristics will only work for "old-style" annotations, but that's OK I think.

### DIFF
--- a/pybv/_export.py
+++ b/pybv/_export.py
@@ -117,18 +117,6 @@ def _mne_annots2pybv_events(raw):
         # defaults to type="Comment" and the full description
         etype = "Comment"
         description = annot["description"]
-        for start in ["Stimulus/S", "Response/R", "Comment/"]:
-            if description.startswith(start):
-                etype = start.split("/")[0]
-                description = description.replace(start, "")
-                break
-
-        if etype in ["Stimulus", "Response"] and description.strip().isdigit():
-            description = int(description.strip())
-        else:
-            # if cannot convert to int, we must use this as "Comment"
-            etype = "Comment"
-
         event_dict = dict(
             onset=onset,  # in samples
             duration=duration,  # in samples
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -424,7 +424,7 @@ def _chk_events(events, ch_names, n_times):
 
         for event in events_out:
             if event["description"] < 0:
-                raise ValueError(f"events: descriptions must be non-negative ints.")
+                raise ValueError("events: descriptions must be non-negative ints.")
             tformat = event["type"][0] + "{:>" + str(twidth) + "}"
             event["description"] = tformat.format(event["description"])
 
--- a/pybv/io.py
+++ b/pybv/io.py
@@ -89,15 +89,10 @@ def write_brainvision(
                 dimension of the `data` array.
             - ``"duration"`` : int
                 The duration of the event in samples (defaults to ``1``).
-            - ``"description"`` : str | int
-                The description of the event. Must be a non-negative int when `type`
-                (see below) is either ``"Stimulus"`` or ``"Response"``, and may be a str
-                when `type` is ``"Comment"``.
+            - ``"description"`` : str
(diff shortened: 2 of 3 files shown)


sappelhoff:
Thanks Clemens, I will have time to review this later this week!

cbrnr:
Here's something to consider: pybv can already export arbitrary markers as type `Comment`, so maybe there is no need for such a big change? Coming from MNE, the private `_export_mne_raw()` function already does the job (it sets the type of all markers to `Comment`).

pybv/_export.py
```diff
@@ -117,18 +117,6 @@ def _mne_annots2pybv_events(raw):
...
-                etype = start.split("/")[0]
-                description = description.replace(start, "")
-                break
-
-        if etype in ["Stimulus", "Response"] and description.strip().isdigit():
-            description = int(description.strip())
-        else:
-            # if cannot convert to int, we must use this as "Comment"
-            etype = "Comment"
-
```
sappelhoff, line 131: I think removing this would result in some lines that are not pretty, like:

`Mk2=Comment,Stimulus/S253,487,0,0`

when it could also (more sensibly, as currently done) be:

`Mk2=Stimulus,S253,487,0,0`

pybv/io.py
```diff
@@ -151,11 +146,6 @@ def write_brainvision(
     channels with non-voltage units such as °C as is (without scaling). For maximum
     compatibility, all signals should be written as µV.
 
-    When passing a list of dict to `events`, the event ``type`` that can be passed is
-    currently limited to one of ``{"Stimulus", "Response", "Comment"}``. The BrainVision
-    specification itself does not limit event types, and future extensions of ``pybv``
-    may permit additional or even arbitrary event types.
```
sappelhoff, line 157: Fine to remove this, but I would somewhere keep a documentation that "Stimulus", "Response", and "Comment" are the classic types in a BrainVision file that is written by BrainVision Recorder.

sappelhoff:
>  Coming from MNE, the private _export_mne_raw() function already does the job (it sets the type of all markers to Comment).

I assume you are referring to this?

https://github.com/bids-standard/pybv/blob/79dceaa02f6a2da77e8d5b27979ee3f378626137/pybv/_export.py#L126-L130

> Here's something to consider: pybv can already export arbitrary markers as type Comment, so maybe there is no need for such a big change?

I personally am able to write all data that I want to write. I thought your argument in #103 convincing so I am fine if you really need this change. But if the changes here are not a big advantage for you, then we should just leave it as is IMHO.

cbrnr:
Since MNE recently added better support for importing BV annotations (https://github.com/mne-tools/mne-python/pull/12652 and https://github.com/mne-tools/mne-python/pull/13100), the current export functionality is sufficient also for my use cases. I am therefore closing this PR.

PR #121, opened 2024-08-27
DOC - point to FieldTrip public API rather than private functions 

robertoostenveld:
This updates the README to point to the the end-user API functions in the `fieldtrip/fileio` directory rather than the private functions which are not to be called directly


sappelhoff (approved):
Thanks @robertoostenveld

PR #127, opened 2025-02-18
Fix test for MNE-Python main

cbrnr:
MNE-Python now discards the first "New Segment" annotation automatically, since it only contains the recording time (see https://github.com/mne-tools/mne-python/pull/13100).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ test = [
   "build",
   "matplotlib",
   "mne",
+  "packaging",
   "pre-commit",
   "pytest",
   "pytest-cov",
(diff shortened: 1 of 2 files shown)


cbrnr:
I added `packaging` (to the `test` extras group) to check the MNE version, but this is a third-party package, so let me know if that's OK or not.

sappelhoff (commented):
ok with me to add "packaging" to the test deps, however something still seems to be off according to the CI.

cbrnr:
@sappelhoff all green now!

sappelhoff (approved):

sappelhoff:
Thanks!

PR #128, opened 2025-02-18
Support Python 3.13 and drop 3.9

cbrnr:
That's it.

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.12"]
+        python-version: ["3.13"]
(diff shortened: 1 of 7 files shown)


sappelhoff (dismissed):
thanks! you could add a chnagelog entry for it, if you wanted to.

cbrnr:
I will, and let's wait until #127 is in.

sappelhoff (approved):

cbrnr:
Is it normal that these four jobs take forever to start?

sappelhoff:
> 
Is it normal that these four jobs take forever to start?

nope, I just realized that when we change the CI runners (no 3.9 anymore, no 3.12 anymore, but 3.13) then we also need to change the config of which checks are expected.

Will do that now, that is something to do in the repo settings.

